PR #6, opened 2023-09-27
fixing links across markdown documents

miki725:
also all links are validated via pre-commit as manually keeping up with them is not very effective...


viega (approved):
Perfect, thank you.

PR #8, opened 2023-09-28
adding github actions for lint/tests

miki725:
this adds automatic github actions for running linting/test checks on open PRs.

linting checks are validated by https://pre-commit.com/

other tests are using https://github.com/pytest-dev/pytest

---

tests only run when there are any source code changes (e.g. changes to `.nim` files). to test PR tests I temporarily enabled running them even without any source code changes (as this PR doesnt change any source files). that test run can be seen on:

https://github.com/crashappsec/chalk/actions/runs/6342089679/job/17227124600?pr=8

tests pass 🎉

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,34 @@
+name: lint
+
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - ready_for_review
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+
+    if: |
+      !github.event.pull_request.draft &&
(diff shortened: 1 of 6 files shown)


viega (approved):
Noting for anyone external who reads the comments that this is enabling running the tests currently available via `make test`, which we would have liked to gotten in before opening. But these are the same tests we've been running internally, and that are available when running `make tests` now.

@miki725 I assume this is going to add a status badge, yes? I don't see that declaratively anywhere in here; I may be missing it, or maybe it's set elsewhere since we're using gh actions?

Either way, if that's happening, this all is good on my end.

viega (approved):
Thanks for the changes, looks good, sir!

PR #9, opened 2023-09-29
Minimize use of tmpfiles when storing Dockerfiles

viega:
Generally, systems use /tmp for temporary files, and most modern API interfaces to using /tmp take mitigation against file-based race conditions, for instance, by leveraging per-app directories and randomness in selecting file names.

However, there are times when the system default isn't a good option for Chalk when it needs temporary space. Specifically, we've learned that, for those running Docker via Snap on Ubuntu systems, Snap's isolation of temporary files means that users will get an error if we try to use /tmp to, for instance, write out a temporary docker file that we want to use with a container.

To mitigate that problem, the primary use of tmp directories in the docker path is for writing out a temporary docker file whenever additions are requested. With this PR, when we need to extend the docker file, we pass it over stdin if it's at all possible.

The only case where that's NOT possible is if the docker invocation is already using stdin to pass a context. That doesn't seem to be common, but in those cases, we still use a temporary file.

However, it's possible to address the problem externally to chalk by setting the TMPDIR environment variable; the underlying tmp file interface respects it, and it's a standard variable.

However, Chalk philosophically doesn't want to leave opportunity for people to "forget" to do things when deploying us. So we also added the `default_tmp_dir` field for the configuration, which allows you to pick a place for temporary files to use IF no value for TMPDIR is provided.

We assume if the user provides one, since it's so standard, they know what they're doing. For that reason, we didn't add a command-line flag; there's already a paradigm for this.

If a user does not use this variable and `TMPDIR` is not set, they will probably end up with `/tmp` or `/var/tmp`, which should be great in most cases now that dockerfiles aren't generally going to a tmp file.

### DIFF
--- a/chalk.nimble
+++ b/chalk.nimble
@@ -8,7 +8,7 @@ bin           = @["chalk"]
 
 # Dependencies
 requires "nim >= 2.0.0"
-requires "https://github.com/crashappsec/con4m == 0.1.5"
+requires "https://github.com/crashappsec/con4m#0040a5c15f7f98ef96046e8b998041760c849bcb"
 requires "https://github.com/viega/zippy == 0.10.7"
 requires "https://github.com/aruZeta/QRgen == 3.0.0"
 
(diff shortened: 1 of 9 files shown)


miki725 (approved):
👍

PR #13, opened 2023-10-01
[enhancement] Remove most command aliases

### LINKED ISSUE #12: consider being less aggressive with command aliasing
Supporting short command aliases means that the user can type less. But it also complicates adding commands that share a prefix with already-existing commands.

For example:

```console
$ chalk --no-color help insert | grep 'ins,'
  •  a̲l̲i̲a̲s̲e̲s̲ ̲ ̲: inject, ins, in, i
```

means that people will inevitably write scripts with `chalk i` ([Hyrum's Law](https://www.hyrumslaw.com/)), and get used to writing `chalk i`. So if chalk adds another command that begins with `i`, chalk must either make `chalk i`:

- An error, saying that it's ambiguous. I'd probably argue that this is a breaking change that requires a major version bump if chalk uses [semantic versioning](https://semver.org/), especially if chalk has not explicitly stated that a chalk update may break aliases.
- Work as before, but then the mapping of aliases to commands becomes complicated (`chalk f` continues to perform `chalk foo`, even when `f` is no longer an unambiguous prefix after `chalk frobnicate` is added).

Solutions:

1. State the contract somewhere. For example:
   - "a chalk update will never break an alias"
   - or "chalk makes no guarantees about aliases - scripts that use chalk should always use full chalk command names"
1. Alias less aggressively.

I would suggest the strong form of option 2. In my experience, aliases in CLI tools are not worth the increased complexity for maintenance and backward compatibility. I would suggest supporting only full command names, and delegating keystroke-saving to tab completion. `chalk i[Tab]` is only one keystroke more than `chalk i`.

Also, the current implementation of command aliasing does not support any unambiguous prefix. This makes it hard to remember which aliases are supported. For example, `chalk vers` is an alias for `chalk version`, but `chalk ver` isn't.

In other tools:

```console
$ git check
git: 'check' is not a git command. See 'git --help'.

The most similar command is
	checkout
```

```console
$ docker hist
docker: 'hist' is not a docker command.
See 'docker --help'
```

```console
$ gh var
unknown command "var" for "gh"

Did you mean this?
	pr
	variable
```

Chalk version: https://github.com/crashappsec/chalk/commit/ea36ce1c185731bba9ba191bc1a209a1da2bf6c1

viega:
Ticket #12 is correct in my view that the command aliasing is not necessary and probably not even a good idea, considering we're auto-installing a bash completion script.

This PR goes ahead and removes most aliases from the command line configuration file. The only two cases are `inject` as an alias for `insert`, and `environment` for `env`, because I do still prefer the shortened default (we've been calling it the `env` command for a long time, but people should still be able to type the whole thing out if they want).

Closes #12


viega:
None of the tests used aliases, so everything is passing locally, so took it out of draft.

nettrino (approved):

PR #34, opened 2023-10-04
[fix] Enable (most) wrapping when seeing USER directive

### LINKED ISSUE #33: Wrapping is currently incompatible with the USER directive
## Description

Proxying for people who have been using the tool  who saw this in a few of their builds.

When entry point wrapping is on, it currently won't succeed if the `USER` directive is present, full stop.

Apparently, once a `USER` directive appears in a Dockerfile, `COPY` still happens with root permissions. So copying in the chalk binary to be the new entry point works, but our attempt to set the perms will NOT work, since the file is owned by root.  This can, of course, happen if the user is set in a base image.

We made the conscious choice to do the `COPY ...` and `RUN chmod ...` to add +x because the `--chmod` flag to `COPY`:

a) Requires BUILDKIT to be on, and we've heard that plenty of people still turn it off.
b) Only showed up in v20... 3 years isn't a long time in some places.

BUT, properly keeping track of the user would be a PITA, if it's not explicit in the last section of the Dockerfile.

I think what we are going to do for this:

1) IF buildkit is on and v20 or higher is available, we will just always use `--chmod`.
2) IF not, and there is a `USER` directive after the last `FROM`, we will do what we're doing today, but sandwich our work with `USER root` and a repeat of their `USER` directive.
3) If the `USER` directive is NOT found, or if there's one before the last FROM statement, we're just going to keep with the current behavior, instead of doing something too kludgy.  For people who explicitly disable buildkit, we'd be able to give them some options here, but I think building anything more into the software at this point doesn't make sense given:

## Impact

Not a lot of images use the directive. Since root in a container doesn't generally mean root, most people just don't change it. 

With a fix, people who really want to be conservative and chalk everything could force no wrapping when buildkit is off, or force buildkit on.

## Steps to Reproduce

with
```docker.wrap_entrypoint = true```

Anything like:

```
FROM alpine as base

RUN adduser -D testuser
RUN id
USER testuser
RUN id
```

Will result in Chalk taking the _failsafe_ path, where it reverts to running the original `docker` command without any chalk mark being added, or entry point wrapping occuring.

viega:

## Issue

Closes #33 

## Description

Implements changes discussed in #33 that make wrapping in the face of a USER directory work in most cases. In particular:

1) Whenever Buildkit is used and the docker version is 20 or higher, use the --chmod flag to COPY.

2) If # 1 doesn't apply, and the last section has a USER directive, then we switch the USER to root before the copy/chmod, then we switch it back at the end.

If neither apply and there is a user inherited from a base image, then wrapping will generally still fail.

## Testing

This is still in draft b/c there's no test yet. Ideally, we would do all three of:

1) Create a basic container with a USER in the last section, see if it works. Should build via chalk.
2) Use the same Dockerfile, but try it with BUILDKIT disabled. Should build via chalk.
3) Create a two-section dockerfile,  with the USER directive in the first section. Should kick to the failsafe path.


miki725 (approved):
👍 this has edge cases however this is prolly good enough for most people as buildkit is pretty popular and we documented current limitation in release notes

PR #37, opened 2023-10-05
test(gdb): adding gdb tests for elf encoding (on behalf of Liming)

miki725:
@indecisivedragon  did most of the work for the gdb stuff so she deserves the credit! (Im just syncing between repos here for the most part)

gdb tests check some basic things about elf encoding by:

* running binary before chalking it and reporting:
  * couple of dozen bytes after the program entrypoint as seen in the gdb breakpoint
* chalking the binary
* running same gdb script on the chalked binary and comparing the report before chalk and after. If all goes well, there should not be any difference in the instructions of the entrypoint comparing chalked and non-chalked binary.

In the future we can add more checks here as whatever json gdb script outputs to stdout test will automatically compare.

Speaking of json, some of the bits for looking for valid json in the os.py/runner.py needed to be changed so that the json extraction logic can work for both chalk reports as well as other commands like gdb in this case.

## Description

ELF encoding was not directly tested before so it could of been producing non-runnable binaries without tests picking that up. With gdb tests, it ensures encoded ELF binary is runnable after being chalked.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,3 @@
-*config-tool*
 .dockerignore
 .github
 Dockerfile*
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,4 @@
 .TODO
-.config-tool-bin/
-config-tool-outdir
 /bin/chalk
 /chalk
 /src/getlibpath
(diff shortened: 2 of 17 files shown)


indecisivedragon (dismissed):
nothing could possibly be wrong with any of this code

viega (approved):

PR #43, opened 2023-10-06
feat(docker): adding METADATA_ID/_REPO_{TAGS,DIGESTS} to push outputs

miki725:

## Issue

`docker push` didnt by default sent `METADATA_ID` which made it very hard to link built image to pushed image

## Description

by default adding these keys to push templates as they are important to link built image with pushed image so they should be enabled by default

in addition adding specific test which tests both
* docker build && docker push
* docker build --push and asserts that required keys are present in both reports

misc: also removing unused configs from tests/data

## Testing

```
make tests args="test_docker.py::test_build_and_push"
```


src/configs/base_report_templates.c4m
```diff
@@ -1662,6 +1663,8 @@ running insert commands.
   key._VIRTUAL.use                            = true
   key._CURRENT_HASH.use                       = true
   key._IMAGE_ID.use                           = true
+  key._REPO_TAGS.use                          = true
```
viega, line 1666: Not sure these should be in the audit report? What was your logic there?

src/configs/base_sinks.c4m
```diff
@@ -165,6 +165,7 @@ configuration:
 
 ```
 sink_config my_https_config {
+  enabled: true
```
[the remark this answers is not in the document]
miki725: it was in other docs so just adding to match

src/configs/base_report_templates.c4m
miki725: I think github shows diff badly here. this is actually inside `report_template terminal_insert {`

viega (approved):

PR #45, opened 2023-10-06
fix(heartbeat): fixing flaky heartbeat test due to invalid probe build tag

miki725:

## Issue

heartbeat test is flaky

## Description

The root cause was that probe build to get current build platform, docker tag name was random which sometimes resulted in invalid tag names:

```
ERROR: invalid tag ".mpiljfkim.ahsfgpf3f4g": invalid reference format
```

docker tag names need to start/end with alphanumeric character so to solve it adding `chalk_` and `_probe` prefix/suffix. This ensures the build tag is correct and docker can build it.

## Testing

```sh
until not 'make tests args="test_docker.py::test_docker_heartbeat --slow --pdb"'
```

### DIFF
--- a/src/commands/cmd_docker.nim
+++ b/src/commands/cmd_docker.nim
@@ -202,26 +202,12 @@ proc getDefaultPlatformInfo(ctx: DockerInvocation): string =
     return ctx.foundPlatform
 
   let
-    probeFile    = """
+    probeFile      = """
 FROM alpine
 ARG TARGETPLATFORM
 RUN echo "CHALK_TARGET_PLATFORM=$TARGETPLATFORM"
 """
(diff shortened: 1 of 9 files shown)


viega:
Looked at it w/ Liming; the issue is that the docker invocation in the wrap probe needs `--no-cache`

miki725:
initially I thought so as well but that doesnt seem to be the case. even with cache it still shows the platform in the `RUN` command itself:

image

we are getting very close though to the root cause 🤞

miki725:
ok so the root cause was related to invalid probe docker image build tag name. see updated PR description

indecisivedragon (approved):

viega (approved):

PR #47, opened 2023-10-09
feat(config): Load and configure components without editing configs directly

### LINKED ISSUE #31: Component system for config files
## Description

This would implement a component system for Chalk, where one could incorporate and configure small changes to a configuration, without having to worry about conflicts with other components, and without having to actually directly edit the configuration file.

One would just have to provide values for parameters denoted by the component, such as the IP address if setting up reporting to an HTTPS endpoint.

Details for this are described in the design doc (#30).

## Result

See #30 

## Dependencies

N/A; note that most of the work will happen in the con4m repo.

## Related Tickets

Design Doc: #30 
Objective: #21 (command line usability)

viega:
Initial (untested) integration of new Con4m component system.

## Issue

Closes #31 

## Description

Implements the initial component system that allows people to try out examples simply via `chalk load`.

See documentation here: https://github.com/crashappsec/chalk/discussions/30

## Testing

Let's build out as many real configurations as practical that we'd actually like to provide by default, and then see how they run!


viega:
This is probably ready (and big enough) for others to start tire kicking. Once I update the design doc, I'll update the ticket description and move this out of draft.

viega:
Seems like virtually all tests fail here. That's surprising; I was expecting things around `chalk load` behavior might fail, but I wouldn't have expected anything else to fail. I'll be looking at that tomorrow.

viega:
My initial guess is that, since most of the tests are loading a config, there's some unintended consequence that I wasn't thinking about.

Come to think about it, the tests definitely need to start passing `--replace` to avoid being treated like component, so that will be my first order of business.

drraid (approved):

PR #48, opened 2023-10-10
docs: improve developer doc for testing

indecisivedragon:

## Issue
closes #https://github.com/crashappsec/chalk-internal/issues/835

## Description
This updates the README.md in tests to be more informative and useful.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -105,7 +105,11 @@ Any arguments passed in through `args` will be directly passed through to the un
 
 ### Running Tests Directly
 
-While tests can be run directly via `pytest`, this is not recommended. However, if you would like to do so, ensure that you have the following:
+While tests can be run directly via `pytest`, this is not recommended; they are intended to be run through the `docker compose` environment, and many of them are likely to fail without the correct setup.
+
+WARNING: These tests are ELF only.
+
+If you would really like to run tests directly, ensure that you have the following:
 
 - `python` version 3.11 or greater
 - `pipx` (recommended) or `pip` package installer for python
@@ -114,7 +118,8 @@ While tests can be run directly via `pytest`, this is not recommended. However, 
 To set up the testing framework:
 
 1. Install poetry with `pipx install poetry`
-2. Install dependencies with `poetry install` (the list of dependencies to be installed is located at `tests/pyproject.toml`)
+2. From the repository root: `cd ./tests`
+3. Install dependencies with `poetry install` (the list of dependencies to be installed is located at `tests/pyproject.toml`)
 
 To run the tests:
 
--- a/tests/README.md
+++ b/tests/README.md
@@ -124,8 +124,9 @@ To set up the testing framework:
 To run the tests:
 
 1. From the repository root: `cd ./tests`
-2. Start poetry shell with `poetry shell`
-3. Run tests with `pytest` (flags and arguments as in the previous section)
+2. Run tests via poetry with `poetry run pytest` (flags and arguments as in the previous section)
+
+Alternatively, you can start a poetry shell with `poetry shell` and then run `pytest`, but this may not work for all shells.
 
 WARNING: Since the chalk tests were intended to be run via `docker compose`, running them directly through `pytest` will cause a number of failures. In particular:
 
--- a/tests/README.md
+++ b/tests/README.md
@@ -1,8 +1,8 @@
 # Chalk Testing
 
-This directory contains all the basic functionality tests for chalk. Tests are run via `docker compose` in the `tests` container which internally uses the `pytest` framework.
+This directory contains all the basic functionality tests for chalk. Tests are run via `make tests`, which `docker compose` in the `tests` container which internally uses the `pytest` framework.
 
-While `pytest` can be used to run the tests directly,
+While `pytest` can be used to run the tests directly, it is recommended to use the makefile
 
 ## Setup
 
--- a/tests/README.md
+++ b/tests/README.md
@@ -1,10 +1,10 @@
 # Chalk Testing
 
-This directory contains all the basic functionality tests for chalk.
-Tests are run via docker compose in a separate `tests` container which
-internally uses the `pytest` framework.
+This directory contains all the basic functionality tests for chalk. Tests are run via `docker compose` in the `tests` container which internally uses the `pytest` framework.
 
-## Requirements
+While `pytest` can be used to run the tests directly,
+
+## Setup
 
 ### Chalk Binary
 
@@ -38,7 +38,7 @@ Please ensure you have an up-to-date container with:
 docker compose build tests
 ```
 
-### Running Tests
+## Running Tests
 
 All commands given are assumed to be run from the root of the repo.
 
@@ -117,7 +117,7 @@ make tests args="[TESTFILE]::[TESTNAME] --pdb"
 Alternatively you can add `breakpoint()` before the failing assertion
 and manually invoke the single test.
 
-### Adding a Test
+## Adding a Test
 
 #### Pytest Convention
 
--- a/tests/README.md
+++ b/tests/README.md
@@ -56,7 +56,9 @@ WARNING: Debug builds are very slow, so it is not recommended to run the entire 
 
 ## Running Tests
 
-All commands given are assumed to be run from the root of the repo.
+### Makefile
+
+The easiest way to run tests is via the makefile. All commands given are assumed to be run from the root of the repo.
 
 To run all tests:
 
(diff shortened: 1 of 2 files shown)


tests/README.md
```diff
@@ -1,44 +1,65 @@
...
-1. Compile `chalk`:
+1. Build chalk deps container:
 
-   ```sh
-   # root of the repo
-   make chalk
-   ```
+```sh
+docker compose build chalk
+```
```
miki725, line 38: you can indent a block in markdown so that it will render inside the numbered list. In this case it will need to be 3 spaces indent to match `1. `

tests/README.md
```diff
@@ -1,44 +1,65 @@
...
 
-The second command should drop a `chalk` binary that is usable by tests.
+(this step can be skipped if the `chalk` container is up to date)
```
miki725, line 40: same for rest of the text block inside lists. if you indent it, they will show up as being inside list. for example:

```
1. foo
   ```sh
   nested code
   ``
    description
2. bar
```

will render as:

1. foo
   ```sh
   nested code
   ```
    description
2. bar

tests/README.md
```diff
@@ -70,28 +91,35 @@ make tests args="[TESTFILE]::[TESTNAME]"
...
-See [pytest docs](https://docs.pytest.org/en/7.1.x/how-to/usage.html)
-for more invocation options.
+To run a single case of a single test:
+
+```sh
+# root of the repo
+make tests args="[TESTFILE]::[TESTNAME][test-case]"
+```
+
+ex: `make tests args="test_elf.py::test_virtual_valid[copy_files0]"`
```
miki725, line 101: can we move example into code block above. it will render better vs being inlined

tests/README.md
```diff
@@ -114,42 +144,48 @@ To do that run test(s) with `--pdb` flag:
 make tests args="[TESTFILE]::[TESTNAME] --pdb"
 ```
 
-Alternatively you can add `breakpoint()` before the failing assertion
-and manually invoke the single test.
+Alternatively you can add `breakpoint()` before the failing assertion and manually invoke the single test.
+
+### Container Shell
+
+You can also drop into a shell in the ` tests` container via `docker compose run --entrypoint=sh tests`, and from there run `pytest` manually.
```
miki725, line 151: should we add `make tests/sh` or something to fasciliate that? I think you need to do that pretty often

tests/README.md
```diff
@@ -114,42 +144,48 @@ To do that run test(s) with `--pdb` flag:
...
+
+All python tests must follow `pytest` conventions to be picked up by the test runner:
+
+- New test files should be added to the `tests/` directory as `test_filename.py`. A new test file should be created to test new functionality, for example if support has been added for a new codec.
+- Individual tests should be added to a test file as a new function as `test_functionname`. Each individual test in a test file should test some aspect of that test file's functionality; if the test you want to add doesn't fit, consider if it would be more appropriate to create a new test file and add it there instead.
+- A new test case for an existing test can be added via the pytest `paramatrize` fixture, if appropriate.
+
+### Datafile Location
+
+All new test files should added to the `tests/` directory, and any test data should be added to the `tests/data` directory.
```
miki725, line 165:
```suggestion
All new test files should be added to the `tests/` directory, and any test data should be added to the `tests/data` directory.
```

tests/README.md
```diff
@@ -114,42 +144,48 @@ To do that run test(s) with `--pdb` flag:
...
 
-### Adding a Test
+WARNING: Any files (including test files and data files) that are NOT in the `/tests` directory will not be accessible from within the `tests` container. Any data files that need to be in a specific path for testing (ex: config files loaded from `/etc/chalk`) must be copied to the target path as part of test setup, which happens inside the container after startup.
```
miki725, line 167: thats not true. the whole repo is actually mounted inside container. by convention we just copy things to `tests/data` but some tests do access things outside of it. for example the nimble version check or outconf tests which load raw configs from the nim source

tests/README.md
```diff
@@ -114,42 +144,48 @@ To do that run test(s) with `--pdb` flag:
...
+### Test Fixtures
 
-All python tests must follow `pytest` conventions to be picked up by the
-test runner.
+Global test fixtures are defined in `conftest.py`. Any test fixtures used across multiple test files should be defined in `conftest.py`; any test fixtures used only in a single test file should be defined in the test file.
 
-- New test files should be added to the `tests/` directory as
-  `test_filename.py`.
-- Individual tests within the test file should be named `test_testname`.
+More information about fixtures can be found [here](https://docs.pytest.org/en/6.2.x/fixture.html).
```
miki725, line 173: lets use consistent version of pytest for all pytest links. above there are links to 7.x but here its 6.x

tests/README.md
```diff
@@ -114,42 +144,48 @@ To do that run test(s) with `--pdb` flag:
...
 
-All test data should be added to `tests/data` directory.
+- `tmp_data_dir`: creates a temporary data directory for each test that will be destroyed at the end of that test. All tests are run from within their temporary directories, and each test has its own temporary directory that does not conflict with any other temporary directories. It is recommended that any tests that mutate data (ex: chalking a system binary) first copy that data into the `tmp_data_dir` and then act on the copy, so that subsequent or parallel tests don't run into conflicts.
+- `copy_files`: copies files into the temporary directory.
+- `chalk`: retrieves the chalk binary to be used in testing (which will be the chalk binary at the root of the repository), and loads it with the default testing configuration which subscribes to console output (ensuring that we get chalk reports to stdout even if chalk is not run in a tty). Note that the scope of this fixture is `session`, so this will be the SAME chalk binary for ALL tests in a single invocation of `make chalk`. If your test needs to make any changes to the chalk binary itself, use `chalk_copy` instead.
```
miki725, line 179: single invocation of `make tests`... not `make chalk`

tests/README.md
indecisivedragon: I do do it pretty often, a `make` command for this would be great! Let's do it in another pr and update the doc there, I'd like to keep this one just docs without code changes.

tests/README.md
indecisivedragon: does vale catch grammar issues also? :scream_cat:

miki725: some basics I think if you enable appropriate styles (they have defaults for microsoft, google and redhat - https://vale.sh/generator/). this was just me but Im not English expert so I have no idea if its actually correct. my logic if it sounds right, its prolly right but what "right" actually is 🤷

tests/README.md
```diff
@@ -1,17 +1,35 @@
...
-Tests are run via docker compose in a separate `tests` container which
-internally uses the `pytest` framework.
+This directory contains all the basic functionality tests for chalk. Tests are run via `make tests`, which uses `docker compose` to run the `tests` container which internally uses the `pytest` framework.
 
-## Requirements
+While `pytest` can be used to run the tests directly, it is recommended to use the makefile script instead, as that provides a more consistent developer experience.
+
+## Setup
+
+Before running tests, ensure that `docker` and `docker compose` are installed.
```
viega, line 9: Can we list dependencies and give an option for how to run this stuff NOT in docker?  Obviously some Linux-specifiic stuff would break, but:

a) The process does slow things down a lt
b) We probably should run tests suites natively on a Mac anyway.

indecisivedragon: Have added instructions with several caveats.

tests/README.md
```diff
@@ -1,17 +1,35 @@
...
-Tests assume that a chalk binary is present in the root directory of
-the repo, and will use that binary in all tests cases.
+Upon starting a test run, the script will first look for an up-to-date chalk binary in the root directory of the repo. If there is no chalk binary, or if the chalk binary is not up to date (ie, there are local code changes that have not yet been built), the script will rebuild the chalk binary.
+
+WARNING: If there is already a chalk binary in the root directory that the script considers "out of date", this binary will be DELETED. If you want to keep this binary, move or rename it.
+
+Tests will always rebuild the binary if it detects a code change in the underlying nim code. However, you will need to build `chalk` manually when switching between `release` and `debug` builds:
+
+1. By default, the testing script will build a `release` binary for chalk, so if you want to test against the `debug` build instead you must build it yourself.
+2. If you have been working with a `debug` build so far, but want to switch to `release`, you will need to rebuild manually as the test script will not rebuild if there have not been any code changes.
```
viega, line 30: I got the manual instructions running on main, but the tests don't really work there:

========================== 50 failed, 32 passed, 17 skipped, 2 errors in 100.05s (0:01:40) 

Some, maybe all of it, is clearly due to an assumption of ELF. I think that's a bit of an issue. But at the very least it needs to be clear in the docs that they are ELF only.

tests/README.md
```diff
@@ -70,12 +88,54 @@ make tests args="[TESTFILE]::[TESTNAME]"
...
+While tests can be run directly via `pytest`, this is not recommended. However, if you would like to do so, ensure that you have the following:
+
+- `python` version 3.11 or greater
+- `pipx` (recommended) or `pip` package installer for python
+- `docker` for the docker tests
+
+To set up the testing framework:
+
+1. Install poetry with `pipx install poetry`
+2. Install dependencies with `poetry install` (the list of dependencies to be installed is located at `tests/pyproject.toml`)
```
viega, line 117:
```
cd tests
poetry install
```


(Without that, it doesn't work)

viega (approved):
Two small additions then it's fine to merge

tests/README.md
indecisivedragon: have added a warning at the top

tests/README.md
```diff
@@ -70,12 +88,59 @@ make tests args="[TESTFILE]::[TESTNAME]"
...
+
+1. Install poetry with `pipx install poetry`
+2. From the repository root: `cd ./tests`
+3. Install dependencies with `poetry install` (the list of dependencies to be installed is located at `tests/pyproject.toml`)
+
+To run the tests:
+
+1. From the repository root: `cd ./tests`
+2. Start poetry shell with `poetry shell`
+3. Run tests with `pytest` (flags and arguments as in the previous section)
```
miki725, line 128: poetry shell has some implications which might break for some people. in docs it might be simpler to do `poetry run pytest ....` which internally will activate appropriate venv and then run pytest in that env without explicitly activating any of the shell bits (bash/zsh/fish/etc)

miki725 (dismissed):

miki725 (approved):

viega (approved):

Issue #27, opened 2023-10-03
Chalk may appear to hang while silently downloading cosign


ee7:
### Example

For a given binary `foo`, run:

```shell
chalk insert foo
```

with Chalk version 175f32f001d90de02c3efa10d19467f08e98a1a7.

### Expected behavior

Should never appear to hang; should print a status message if some potentially slow operation is happening.

### Observed behavior

May appear to hang when cosign is not found. The cosign binary is 98 MiB (for x86_64 Linux), so there can be significant delay on slower connections, or for transient network issues.

No information is printed, so a user may perform a keyboard interrupt.

### Possible solution

Write to stderr, something like:

```text
cosign not found in PATH, or at '/tmp/cosign'
Downloading cosign... success
```

Bonus: print download progress.

### Other details

When there is no network connectivity, chalk does not appear to hang. It successfully adds the chalk mark, but warns:

```text
warn:  When collecting chalk-time artifact data, plugin implementation metsys threw an exception it didn't
       handle (artifact = /path/to/foo.
```

viega:
Good idea for the interim.  I was really hoping to not be dependent on the cosign tool. We're not using it to do anything too crazy, and I did start on some of the necessary work (the key generation), but also need to implement secret box on top of OpenSSL 3, which we really haven't found time to do.

viega:
Cool we'll check it out.

MyNameIsMeerkat:
This has been initially addressed in PR #49 by fixing an issue that caused `cosign` to be downloaded way more often than intended + additional user messaging. 

Support for `minisign` has been prototyped in a dev branch alongside the abstraction of signing primitives in general, but in doing so a lot of larger signature strategy questions were raised that need further discussion before allocating dev cycles.

More to come on this in future but for the time being the changes #49 will hopefully improve the user experience for those making use of the signing features.

PR #49, opened 2023-10-10
fix(attestation): change default to NOT download cosign outside of setup flow

MyNameIsMeerkat:
- [ X ] Followed the steps in the contributor's guide: https://crashoverride.com/docs/other/contributing#filing-the-pull-request
- [ X ] PR title uses [semantic commit messages](https://nitayneeman.com/posts/understanding-semantic-commit-messages-using-git-and-angular/#fix)
- [ X ] Filled out the template to a useful degree

## Issue

Addressing issue raised in #27 

## Description

Chalk no longer downloads cosign by default outside of the `chalk setup` flow. Previously `plugins/system.nim`'s lookup of the location of the cosign binary caused cosign to be downloaded if it's path could not be determined.  The default behaviour is now that lookups of cosign location will no longer initiate a download unless specifically requested. The `chalk setup` flow will still download cosign as part of its setup process if it cannot be found locally.

## Testing

Run any chalk operation with --trace enabled and see that no cosign download takes place

Run `chalk --trace setup` with no local cosign and  see that a download takes place


src/attestation.nim
```diff
@@ -217,9 +217,13 @@ proc loadFromSecretManager*(prkey: string, apikey: string): bool =
 
   return true
 
-proc getCosignLocation*(): string =
+proc getCosignLocation*(downloadCosign = false): string =
   once:
-    cosignLoc = unpack[string](runCallback(cosignLoader, @[]).get())
+    var args = @[pack(false)]
+    if downloadCosign == true:
+      args = @[pack(true)]
```
viega, line 224: Just a nit here, I'd probably just do:
```
let args = @[pack(downloadCosign)]
```

viega: I think it makes it much clearly what's going on, in its simplicity.

src/configs/attestation.c4m
```diff
@@ -54,12 +56,12 @@ func install_cosign() {
   return false
 }
 
-func load_attestation_binary() {
+func load_attestation_binary(download_if_not_present: bool) {
```
viega, line 59: The `: bool` there should be redundant; I'd expect it to be inferred from the body since the body is unambiguous, but let me know if it does break (it certainly is fine to be explicit about the type here, I don't
think this needs to change).

viega (approved):
Worked great for me, I'd probably ask you to clean up per my first comment there, but otherwise it's good to go and I don't need to look at it again.

src/attestation.nim
MyNameIsMeerkat: Yup agreed, makes sense

viega (approved):
Per previous review.

PR #58, opened 2023-10-15
feat(help) Finished rendering and styled multiple help pages

### LINKED ISSUE #32: Finish the basic functionality for document rendering
## Description

The formatting for help documentation was cut short to get a release out, so whenever output is formatted in any way, everything prints, but it usually does not look good.

The goal was to lay out the content in one pass for sizing (wrapping, alignment), but defer most of the actual stylizing (colors, italics) separately when actually outputting, to avoid most issues with wrapping and overflow (it could never really remove all; terminal resizing is async, and some fonts are not going to give us what we expect, so our character counts could be off).

As part of this work, the html "class" and "id" tags should be respected, so we can fine-tune any thing that doesn't look good.

Additionally, the spacing and default styles need to be set; that was never properly considered; what's there was a placeholder.

## Result

Formatted output shouldn't be embarrassing!

## Dependencies

None, but the work will be almost completely done in nimutils.

## Subtickets

Objectives: #20 #21

viega:

## Issue

Closes #32 

## Description

We released without the renderer being finished, and so there were multiple display issues with displaying help content, particularly in tables. Also things like, if `NO_COLOR` is set, we don't just skip color, we skip any terminal code.

Now, the initial renderer has been finished, and lots of 'features' have been added, for instance, to allow us to have more flexibility in table layout.

There's a default chalk 'style' too, but it's set up to make it easy for us to enable others to configure their own style.

The renderer is 'good enough' now that we will have a very easy time moving to any sort of full-screen model instead of a scrolling model... which I do intend to do soon with the component system.

However, now that things seem pretty healthy, we do need to go back and adjust the docs to format properly. I've seen there are definitely some places where tweaks need to be made, for instance, if the doc is using markdown in a context where markdown doesn't get processed, etc.

## Testing

Try `chalk help` or `chalk help metadata` and see that they're no longer utterly horrid.


viega:
The failing tests are only failing because of error message changes in Con4m and will be passing as soon as #47 is merged.

MyNameIsMeerkat (approved):
All looks good to me

PR #59, opened 2023-10-16
feat(plugins): added metadata for GCP and Azure of cloud providers

nettrino:
## Description
This PR makes the following changes:

- Queries the GCP and Azure metadata endpoints, and updates the `_GCP_INSTANCE_METADATA` and `_AZURE_INSTANCE_METADATA` endpoints respectively.
- Exposes an `_OP_CLOUD_PROVIDER` runtime key, which currently is set to one of `aws`, `gcp`, `azure`.
- [Experimental] Exposes an `_OP_CLOUD_PROVIDER_SERVICE_TYPE` runtime key, which currently is only set for AWS, and gets one of the following values `aws_eks`, `aws_ecs`, `aws_ec2`.
- [Internal] Renames the imds plugin to `cloudMetadata` plugin to reflect its cross-cloud functionality
- [Internal] Renames the `aws` and `ec2` singletons to `cloud_provider` and `cloud_instance_hw_identifiers` respectively.

## Testing

- Download a chalk binary with the above functionality in a node running in the cloud, and run a `chalk insert` command. 
- Run `tail -1 ~/.local/chalk/chalk.log | jq` and observe that the metadata keys have been populated appropriately


MyNameIsMeerkat (dismissed):
Looks good to me, great to have all this abstraction and primitives in there

tests/test_plugins.py
```diff
@@ -211,3 +212,264 @@ def test_imds(
...
+                    "osType": "Linux",
+                    "placementGroupId": "",
+                    "plan": {"name": "", "product": "", "publisher": ""},
+                    "platformFaultDomain": "0",
+                    "platformUpdateDomain": "0",
+                    "priority": "",
+                    "provider": "Microsoft.Compute",
+                    "publicKeys": [
+                        {
+                            "keyData": "ssh-rsa AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAJQPr4RsDbaJdKPHl2gfCwiWcTRVEu0XlQvsPgdvCH/Io8Im1VfBMamtRhTIEqlEoTaRD8h9ETDQAPg7GUVkg07P3ZgDfFf94KePpxADso7GoqaPsGuL4OQpURa4DQCmf1Jw+kDg0TI1ERYIQoNOGduiS5cuB74A5BxcgW2A52ocVoiINS1tPudZBIvnr8iQXa6BhB5EgUVP0w+pGaOgI4jHga8ThT9weGqzBrtBcyiZ44jfT2Tg/AjI4GuXq14HdFEN0096vk= generated-by-azure",
```
miki725, line 309: there is no need to hardcode full RSA key in tests. you can match the value by regex:

```python
"keyData": re.compile(r"...."),
```

see aws example above how to do that

tests/test_plugins.py
```diff
@@ -211,3 +212,264 @@ def test_imds(
...
+    with tmp_file as fid:
+        fid.write(b"Google")
+    bin_path = copy_files[0]
+    insert = chalk.insert(bin_path, config=CONFIGS / "imds.conf")
+    assert insert.report.contains(
+        {
+            "_OP_CLOUD_PROVIDER": "gcp",
+            "_GCP_INSTANCE_METADATA": {
+                "attributes": {
+                    "ssh-keys": 'test:ecdsa-sha2-nistp256 AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAKgXTiO1+sSWCEsq/bWaLdY= google-ssh {"userName":"test@crashoverride.com","expireOn":"2023-10-14T15:11:57+0000"}\ntest:ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQCvddnbJ/XWxMUPXOsDMNoRHJeaCgwqk6g7UYvrXqogwmJ1WpC1QPuG3mhDjmBOcjINi7TYsozDKZilL2BDu2i6CGC1s2Tokq41lsgnCePNdnYmPcA318PmuMmAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAeT7R92kx google-ssh {"userName":"test@crashoverride.com","expireOn":"2023-10-14T15:12:12+0000"}'
```
miki725, line 398: same here

src/configs/chalk.c42spec
```diff
@@ -1992,27 +1992,27 @@ end users.
...
     # https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/identify_ec2_instances.html
     type:    string
     default: "/sys/devices/virtual/dmi/id/product_uuid"
     hidden:  true
-    doc:     "Path where to check AWS product if running in AWS EC2"
+    doc:     "Path where to check product uuid for cloud nodes"
   }
 }
 
-singleton aws {
```
miki725, line 2007: why the rename?

as we officially add multiple cloud providers I think it makes sense to separate them in independant configs as there could be conflicting values

src/configs/chalk.c42spec
```diff
@@ -1992,27 +1992,27 @@ end users.
...
+singleton cloud_provider {
+  gen_fieldname: "cloudProviderConfig"
+  gen_typename:  "CloudProviderConfig"
   gen_setters:   false
   user_def_ok:   false
   shortdoc: """
-Configuration information for the AWS Cloud Provider
+Configuration information for the different Cloud Provider
 """
-  allow ec2
```
miki725, line 2015: similar here. I think `ec2` is clear in aws context but `cloud_instance_hw_identifiers` I dont know what that means without looking at all the keys

src/configs/chalk.c42spec
```diff
@@ -1992,27 +1992,27 @@ end users.
     type:    string
     default: "/sys/class/dmi/id/board_vendor"
     hidden:  true
-    doc:     "Path where to check AWS board vendor if running in AWS EC2"
+    doc:     "Path where to check AWS board vendor if running in AWS, Google or Microsoft nodes"
```
miki725, line 1995: that particular value is set by AWS. see link above for the ec2 docs. do GCP/Azure set the same path?

src/configs/chalk.c42spec
```diff
@@ -1992,27 +1992,27 @@ end users.
...
   }
 
   field sys_product_path {
     # https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/identify_ec2_instances.html
     type:    string
     default: "/sys/devices/virtual/dmi/id/product_uuid"
     hidden:  true
-    doc:     "Path where to check AWS product if running in AWS EC2"
+    doc:     "Path where to check product uuid for cloud nodes"
```
miki725, line 2003: similar here

src/configs/chalk.c42spec
nettrino: the /sys/vendor bits is generic Linux functionality so not tied to ec2

tests/test_plugins.py
nettrino: yeah its a dummy public key anyhow so it was easier to simply paste

src/configs/chalk.c42spec
viega: Personally, I'd prefer to make it super easy to capture things as consistently as possible across multiple clouds, so this approach does work well for me. 

I think if we find something where cloud-specific configuration would be important to people, we can easily support that too, but realistically I can't think of anything anyway.

src/plugins/cloudMetadata.nim
```diff
@@ -0,0 +1,335 @@
...
+    trace("Could not retrieve Azure metadata from: " & $uri)
+    return none(string)
+
+  trace("Retrieved Azure metadata " & uri.path)
+  result = some(response.bodyStream.readAll().strip())
+  if not result.isSome():
+    # log failing keys in trace mode only as some are expected to be absent
+    trace("Got empty Azure metadata from: " & uri.path)
+
+proc hitGcpEndpoint(path: string): Option[string] =
```
viega, line 70: If these are all three the same except for the error message (and perhaps the IP) then should parameterize it instead of keeping 3 copies of the code.

viega (changes requested):

viega (approved):

PR #67, opened 2023-10-19
Jtv/releaseitems

viega:
1. Cleaned up the configurations that can be loaded from `chalkdust.io` and made some minor changes / fixes due to the experience.
2. Made it possible for chalk to auto-update configs for binaries for other architectures that might be used for container wrapping.
3. Release items.


Dockerfile
```diff
@@ -1,88 +1,3 @@
-FROM ghcr.io/crashappsec/nim:ubuntu-2.0.0 as nim
```
miki725, line 1: seems like accidental overwrite?

src/chalk_common.nim
```diff
@@ -356,9 +356,13 @@ var
 
 template dumpExOnDebug*() =
   if chalkConfig != nil and chalkConfig.getChalkDebug():
-    let msg = "Handling exception (msg = " & getCurrentExceptionMsg() & ")\n" &
-      getCurrentException().getStackTrace()
-    publish("debug", msg)
+    let
+      msg = "" # "Handling exception (msg = " & getCurrentExceptionMsg() & ")\n"
```
miki725, line 360: msg empty on purpose? should the commented out code be removed?

miki725 (approved):

PR #86, opened 2023-10-31
feat(docker): adding git context support

miki725:

## Issue

Support for git context is required for default docker push action as it uses git context by default

https://github.com/docker/build-push-action

Otherwise chalk github action cant run with default parameters

## Description

Adding compatibility with docker git contexts. For example:

```
docker build -t chalk http://github.com/crashappsec/chalk.git
```

This is done by:

1. Doing shallow fetch of git context into a bare git repo.
    ```
    git fetch --depth=1 origin <ref>
    ```
2. From there we can extract `Dockerfile` with `git show`
3. Adjust Dockerfile as usual and pass it to build via `stdin`
4. If `stdin` is not supported, git context is checked out locally
   to disk and git context is replaced with path on disk.
   NOTE this is only necessary for non-buildx builds
   and it has edge cases as docker binary does more magic to checkout
   git context such as following all sub-modules recursively
   however for now chalk is kept simple by only checking out the
   main git context repo without any recursion.

---

In addition fixing providing files to docker build as buildx version
check was incorrect as it:

* never called getBuildXVersion() so version comparison was against 0
* did not compare versions semantically so 0.8 was always greater than
  0.11 which is incorrect as per semver

To handle semver comparison, instead of using 3rd party lib
such as https://github.com/euantorano/semver.nim
added very simple `semver.nim` with basic comparison operators.

refactor(semver): removing semver dep in favor of very simple local implementation

## Testing

```
➜ make tests args="test_docker.py::test_git_context --logs -x --pdb"
```

tests:--slow

---

TODOs before merge:

- [x] update git metadata to use git context
- [x] fix segfault with bad TMPDIR permissions


src/util.nim
```diff
@@ -434,3 +442,82 @@ template chalkCloseStream*(chalk: ChalkObj) =
...
+    else:
+      result.add(i)
+
+template get[T](data: seq[T], i: int, default: T): T =
+  if len(data) > i:
+    data[i]
+  else:
+    default
+
+iterator zipLongest*[T](data1, data2: seq[T], default: T): (T, T) =
```
viega, line 485: Personally, I think these semi-functional APIs value conciseness over clarity. You'll notice I never use anything like map()/filter()/reduce(). It's fine to stay in, but please try to avoid, or comment if you really think you need it.

viega (commented):
We should sort out the key situation, but otherwise it's "clean as a whistle." Well done again, sir.

src/docker_git.nim
```diff
@@ -0,0 +1,224 @@
...
+  try:
+    result = runCmdGetEverything(gitExeLocation, allArgs.raw())
+  finally:
+    envVars.restore()
+  if strict and result.exitCode != 0:
+    error("Failed to run git " & allArgs.redacted().join(" "))
+    error(strip(result.stdOut & result.stdErr))
+    raise newException(ValueError, "Git failed")
+
+proc getDefaultBranch(git: DockerGitContext): string =
```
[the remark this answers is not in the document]
viega: Additionally, if we do go out with this, then we basically have to wait a long time to deprecate the config vars that ideally won't be needed for very long. Why not just say, if it's not in PATH, make sure it's in your path?

src/semver.nim
```diff
@@ -0,0 +1,85 @@
...
+## (see https://crashoverride.com/docs/chalk)
+##
+import strutils, util
+
+# very simple semver implementation
+# it this does NOT handle full semver spec
+# it only handles basic dot-separated version format
+
+type Version* = ref object
+    parts: seq[int]
```
viega, line 14: Why isn't parts a three-tuple that has default values if you want to allow short version strings?  It would make the comparison more efficient and more clear too (Many people aren't going to get the 'zip' idiom anyway).

src/semver.nim
```diff
@@ -0,0 +1,85 @@
...
+
+proc parseVersion*(version: string): Version =
+  new result
+  result.parts = @[]
+  for i in version.strip(chars={'v', ',', '.'}).split('.'):
```
viega, line 19: This also relies on parseInt to throw an error for malformed stuff, which makes this highly error prone as a parser. If people accidentally pass in the wrong thing, they could get spurious versions, etc.

OR it could blow up on things it probably should accept, like if people do semver but w/ release candidates... we've all seen `v2.1.0-rc1` for instance. Personally I'd be much more precise and explicit when it comes to what is or is not allowed.

src/semver.nim
```diff
@@ -0,0 +1,85 @@
...
+  for (a, b) in zipLongest(self.parts, other.parts, 0):
+    if a != b:
+      return false
+  return true
+
+proc `!=`*(self: Version, other: Version): bool =
+  return not (self == other)
+
+proc `>`*(self: Version, other: Version): bool =
+  for (a, b) in zipLongest(self.parts, other.parts, 0):
```
viega, line 32: Again, if you use a tuple, you can just rely on the default tuple comparison, which will be far faster and far more clear.

For instance, something like:

`type Version = tuple[major: uint, minor: uint, patch: uint]`

And then `>` and '!=' both work right without you even having to overload something. And it allows people to clearly ask for `somevar.major`.

Or if you wanted to track any dash and tag name after, you could do it in an object, then just compare the tuple fields.



 (and, more specifically, declare Version

src/util.nim
viega: Yeah per above, you don't even need this, since you'd be better off keeping 3-tuples anyway.

viega (changes requested):
Some things to consider; we can discuss whether / if some of my nits should be dealt with now or later.

But despite any comments (which are about both clarity and things where a better approach is possible), it's well considered and pretty clear on the whole, so nice job again.

src/util.nim
miki725: removed

src/configs/chalk.c42spec
```diff
@@ -2561,6 +2561,28 @@ Note that, when chalk is invoked with 'docker' as its EXE name, the default IO c
 """
   }
 
+  field git_exe {
```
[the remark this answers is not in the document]
miki725: removed these configs as eventually well want to use libgit anyway. for now it just searches `PATH`

src/docker_base.nim
```diff
@@ -31,55 +31,69 @@ template extractBoxedDockerHash*(value: Box): Box =
...
-      exeOpt   = chalkConfig.getDockerExe()
-
-    if exeOpt.isSome():
-      # prepend on purpose so that docker_exe config
-      # takes precedence over rest of dirs in PATH
-      userPath = @[exeOpt.get()] & userPath
-
-    dockerPathOpt     = findExePath("docker", userPath, ignoreChalkExes = true)
-    dockerExeLocation = dockerPathOpt.get("")
+    dockerExeLocation = findExePath("docker",
```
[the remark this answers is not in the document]
miki725: changed

src/docker_base.nim
```diff
@@ -31,55 +31,69 @@ template extractBoxedDockerHash*(value: Box): Box =
...
 
-proc getDockerVersion*(): string =
+proc getDockerVersion*(): Version =
   once:
-    let (output, exitcode) = execCmdEx(dockerExeLocation & " version")
+    # examples:
+    # Docker version 1.13.0, build 49bf474
+    # Docker version 23.0.0, build e92dd87
+    # Docker version 24.0.6, build ed223bc820
+    let (output, exitcode) = execCmdEx(dockerExeLocation & " --version")
```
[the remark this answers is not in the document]
miki725: switched all usages I could find to use `runCmd*` which normalizes to `runCommand` from nimutils

src/docker_base.nim
```diff
@@ -117,33 +131,41 @@ proc makeFileAvailableToDocker*(ctx:      DockerInvocation,
...
 
-  if haveBuildContextFlag():
+  if supportsBuildContextFlag():
     once:
       trace("Docker injection method: --build-context")
 
     var chmodstr = ""
 
-    if chmod:
+    if chmod or hasUser:
```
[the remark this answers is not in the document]
miki725: if user is set in the dockerfile, `/chalk.json` would not be accessible as it would be set with root permissions hence the need to change its permissions but Ill change to pass the permission mode

src/docker_git.nim
```diff
@@ -0,0 +1,224 @@
...
+proc contextPath*(git: DockerGitContext): string =
+  return joinPath(git.tmpWorkTree, git.subdir).resolvePath()
+
+proc replaceContextArg*(git: DockerGitContext, args: seq[string]): seq[string] =
+  if git.isCheckedOut():
+    return args.replaceItemWith(git.context, git.contextPath())
+  return args
+
+proc run(git: DockerGitContext,
+         args: seq[string],
```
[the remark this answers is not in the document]
miki725: I usually try to match things as well but sometimes forget...

src/docker_git.nim
```diff
@@ -0,0 +1,224 @@
...
+## This file is part of Chalk
+## (see https://crashoverride.com/docs/chalk)
+##
+import base64, strutils, uri
+import config, util, docker_base
+
+proc setGitExeLocation() =
+  once:
+    gitExeLocation = findExePath("git",
+                                 configPath = chalkConfig.getGitExe()).get("")
```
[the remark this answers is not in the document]
miki725: changed. thanks

src/semver.nim
miki725: changed to tuple

src/semver.nim
miki725: changed comparison to use tuple however as in some places it prints original version back keeping it as object so that it can keep the original version name so things like `1.0` dont print as `1.0.0` if the tuple always has 3 versions

viega (approved):

Issue #88, opened 2023-10-31
Collect most recent author and timestamp from git metadata


drraid:
## Description

Chalk collects git metadata associated with build artifacts, but does not capture the author and timestamp of the most recent commit used to build an artifact, which would provide value as human-readable info that consumers can use to reason about when an artifact was built relative to other events (time of bugs or fixes being introduced in repos, time of incidents, etc.). 

## Result

The chalkmark data should include keys `AUTHOR`, `AUTHOR_DATE`, `COMMITTER`, `COMMITTER_DATE`  with the respective extracted values.

PR #89, opened 2023-10-31
Most recent committer, author, and timestamps from local git repo info

drraid:
Happy Halloween, please enjoy this sPoOkY pull request. 
This PR is to add functionality to meet the requirements of #88, adding the keys: `AUTHOR`, `AUTHOR_DATE`, `COMMITTER`, `COMMITTER_DATE`. This pulls either from the sparse on-disk object files, or the packed objects if the most recent commit was not performed on the currently running branch.


src/configs/base_keyspecs.c4m
```diff
@@ -631,6 +631,50 @@ time of chalk mark insertion.
...
+    type:             string
+    standard:         true
+    since:            "0.1.2"
+    shortdoc:         "Author of most recent commit"
+    doc:              """
+The author of the most recent commit in the current branch, which is likely the same as the committer but not always.
+"""
+}
+
+keyspec AUTHOR_DATE {
```
viega, line 645: Should we call it `DATE_AUTHORED` instead?  AUTHOR_DATE feels like my birthday :)

src/plugins/vctlGit.nim
```diff
@@ -8,18 +8,31 @@
...
   ghBranch     = "branch"
   ghRemote     = "remote"
   ghUrl        = "url"
   ghOrigin     = "origin"
   ghLocal      = "local"
-  eBadGitConf  = "Git configuration file is invalid"
+  gitAuthor    = "author "
+  gitCommitter = "committer "
+  gitIdxAll    = "*.idx"
+  gitIdxExt    = ".idx"
```
viega, line 29: This doesn't seem to be used.

src/plugins/vctlGit.nim
```diff
@@ -8,18 +8,31 @@
...
+  gitIdxHeader = "\xff\x7f\x4f\x63\x00\x00\x00\x02"
+  gitObjects   = "objects" & DirSep
+  gitPack      = gitObjects.joinPath("pack")
+  gitPackExt   = ".pack"
```
viega, line 33: Also unused.

src/plugins/vctlGit.nim
```diff
@@ -191,17 +204,126 @@ proc findGitDir(fullpath: string): string =
...
+    commitDate: string
 
   GitInfo = ref object of RootRef
     branchName: Option[string]
     commitId:   Option[string]
     origin:     Option[string]
     vcsDirs:    OrderedTable[string, RepoInfo]
 
+proc getUint32BE(data: string, whence: SomeInteger=0): uint32 =
+  result = cast[ref [uint32]](addr data[whence])[]
```
viega, line 223: The [] around uint32 isn't needed. I also tend to use `ptr` here instead of `ref` though they both work fine.

viega: Also, the following works:
```
import nativesockets
var be = htonl(x)
```

viega (approved):
Looks good, a couple of minor things you could address or no...

src/plugins/vctlGit.nim
drraid: the top 3 are used, and I intended to use the 4th, I'll fix these

drraid: Ahh I think you only meant to point out the bottom one but it includes the lines above it for context (same as your other comment on unused, where the three above it are used)

src/configs/base_keyspecs.c4m
drraid: Happy to change these, would better keys be:
`DATE_AUTHORED` and `DATE_COMMITTED` ?

src/plugins/vctlGit.nim
drraid: for `ref` vs `ptr` I was just using code I already had handy since nim is still a bit foreign for me, i'll change to `ptr`. I figured there was a `htonl` somewhere but digging felt like more work than shifting bits :). I'll change that too

src/configs/base_keyspecs.c4m
viega: Yeah those sound much clearer.

viega (approved):
👍

PR #93, opened 2023-11-10
fix: capture heisenbug

viega:

## Issue

The fact that the C and Nim APIs disagreed on sign for an integer led to process output capture not getting presented... but only in very rare (but reproducible cases). I haven't even figured out why this mattered or why it didn't work yet. But this seems to fix it?

## Description

Pull in a new nimutils that just changes a 'cint' to a 'csize_t'.  Seriously.
## Testing

If all tests pass, the broken case we found seems to be working on multiple boxes.

### DIFF
--- a/chalk.nimble
+++ b/chalk.nimble
@@ -8,7 +8,7 @@ bin           = @["chalk"]
 
 # Dependencies
 requires "nim >= 2.0.0"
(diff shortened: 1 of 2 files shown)


MyNameIsMeerkat (approved):
Tested locally and fixes things for me 👍

Issue #92, opened 2023-11-02
bug: semgrep download incompatible with `--sarif`


indecisivedragon:
## Description

if `run_sast_tools` is enabled, and semgrep is not found, chalk will attempt to download semgrep with `python3 -m pip install semgrep` which will download version 1.46.0. However chalk also tries to run semgrep with the `--sarif` flag, which is not compatible with version 1.46.0: https://github.com/semgrep/semgrep/issues/9091

## Impact

semgrep (and therefore `run_sast_tools`) is unusable at the moment

## Steps to Reproduce

1. load chalk with a config that has `run_sast_tools: true`
2. run chalk on any binary
3. observe that chalk fails (though this doesn't show you the error)
4. manually run the command that chalk runs: `semgrep ci --config=auto --sarif`
5. observe semgrep error:
```
Object of type ErrorType is not JSON serializable
Traceback (most recent call last):
  File "/home/liming/.local/lib/python3.10/site-packages/semgrep/commands/wrapper.py", line 35, in wrapper
    func(*args, **kwargs)
  File "/home/liming/.local/lib/python3.10/site-packages/semgrep/commands/ci.py", line 539, in ci
    output_handler.output(
  File "/home/liming/.local/lib/python3.10/site-packages/semgrep/output.py", line 345, in output
    output = self._build_output()
  File "/home/liming/.local/lib/python3.10/site-packages/semgrep/output.py", line 483, in _build_output
    return self.formatter.output(
  File "/home/liming/.local/lib/python3.10/site-packages/semgrep/formatter/base.py", line 28, in output
    return self.format(
  File "/home/liming/.local/lib/python3.10/site-packages/semgrep/formatter/sarif.py", line 445, in format
    return json.dumps(output_dict, sort_keys=True, indent=2)
  File "/usr/lib/python3.10/json/__init__.py", line 238, in dumps
    **kw).encode(obj)
  File "/usr/lib/python3.10/json/encoder.py", line 201, in encode
    chunks = list(chunks)
  File "/usr/lib/python3.10/json/encoder.py", line 431, in _iterencode
    yield from _iterencode_dict(o, _current_indent_level)
  File "/usr/lib/python3.10/json/encoder.py", line 405, in _iterencode_dict
    yield from chunks
  File "/usr/lib/python3.10/json/encoder.py", line 325, in _iterencode_list
    yield from chunks
  File "/usr/lib/python3.10/json/encoder.py", line 405, in _iterencode_dict
    yield from chunks
[...]

PR #94, opened 2023-11-10
fix: semgrep installation + runtime arguments now work

indecisivedragon:

## Issue
closes https://github.com/crashappsec/chalk/issues/92

## Description
- semgrep download is now managed by python3 venv instead of pip installing to user installtion
- install location is directly passed to `find_semgrep` so semgrep can be found even if it's not in $PATH
- semgrep default run arguments conflicted with each other, metrics is now default `on` to work with `config=auto`

## Testing

tests need to be added to ensure that semgrep run returns correct sast json

### DIFF
--- a/src/configs/sastconfig.c4m
+++ b/src/configs/sastconfig.c4m
@@ -16,7 +16,7 @@ tool semgrep {
                                                    dict[string, string])
   semgrep_config_profile: "auto"
   semgrep_format:         "sarif"
-  semgrep_metrics:        "off"
+  semgrep_metrics:        "on"
   semgrep_other_flags:    ""
   doc: """
 This runs the semgrep static analyizer.  If it doesn't exist in the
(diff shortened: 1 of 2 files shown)


src/configs/sastconfig.c4m
```diff
@@ -26,29 +26,38 @@ You can configure the following fields in the tool.semgrep object:
...
+semgrep_metrics:        Whether to ping semgrep. 'on' or 'off'. Defaults to 'on' to be compatible with config=auto
 """
 }
 
+
+
 func find_semgrep(ignore) {
-  result := find_exe("semgrep", [])
+  semgrep_path := "/tmp/semgrep.env/bin/semgrep"
+  result := find_exe(semgrep_path, [])
```
viega, line 37: find_exe() IIRC searches the $PATH too, even if you call it with a full path. I don't know why we'd prioritize looking in /tmp if it's already installed on the system??  

I guess if we don't like the version installed?

src/configs/sastconfig.c4m
```diff
@@ -26,29 +26,38 @@ You can configure the following fields in the tool.semgrep object:
...
 
   if result == "" {
-    trace("find_semgrep: Unable to find semgrep in $PATH")
+    trace("find_semgrep: Unable to find semgrep at" + result)
```
viega, line 40: The above means this error message isn't complete...

src/configs/sastconfig.c4m
```diff
@@ -26,29 +26,38 @@ You can configure the following fields in the tool.semgrep object:
...
+      return true
+    }
   }
-  return true
+
+  trace("semgrep could not be installed")
+  return false
 }
 
 func get_semgrep_args(artifact) {
```
viega, line 63: Since we're touching this stuff should we make it more configurable?  Tho I guess a semgrep component might be better for that?

viega (approved):

src/configs/sastconfig.c4m
```diff
@@ -26,29 +26,38 @@ You can configure the following fields in the tool.semgrep object:
...
 }
 
 func install_semgrep(ignore) {
-  output, code := system("python3 -m pip install semgrep")
-  if code != 0 {
-    output, code := system("python -m pip install semgrep")
-  }
-  if code != 0 {
-    return false
+  trace("attempting to install semgrep to python3 venv")
```
miki725, line 47: can be attempt to use `pipx` first as thats pretty common for installing python tooling in the py ecosystem. then as a fallback go to manual venv creation

indecisivedragon: have added pipx as preferential download method

src/configs/sastconfig.c4m
indecisivedragon: I think a separate pr for that would be best

src/configs/sastconfig.c4m
indecisivedragon: mixed up the argument order here which should be fixed now, but yes we want to ensure we have the right version

viega (approved):

PR #99, opened 2023-11-22
refactor(nimutils): Adjust to api changes

viega:

## Issue

As I'm working towards some major under-the-hood con4m improvements, I need to keep Chalk aligned with downstream development to avoid more pain later.

## Description

Lots of little changes to adapt to nimutils changes made while workin…g on con4m stuff. file locking now uses flock() and the use of tables / formatting is much simplified. Also am using a less obnoxious default style.

Instead of having to generate HTML for tables, lists, etc, there's now a much more sane API... most HTML tags get their own functions (but not for 1 letter tags; `paragraph()` instead of `p()`, which had some weird conflict). `quickTable()` is a wrapper around `table(), tr(), td(),` etc.
## Testing

Current testing should be sufficient.

### DIFF
--- a/src/sinks.nim
+++ b/src/sinks.nim
@@ -18,7 +18,7 @@ const
                        "log_prefix"    : MsgFilter(logPrefixFilter),
                        "pretty_json"   : MsgFilter(prettyJson),
                        "fix_new_line"  : MsgFilter(fixNewline),
-                       "add_topic"     : MsgFilter(addTopic),
+                       "show_topic"     : MsgFilter(showTopic),
                        "wrap"          : MsgFilter(chalkLogWrap)
                      }.toTable()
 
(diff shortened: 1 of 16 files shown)


src/sinks.nim
```diff
@@ -18,7 +18,7 @@ const
                        "log_prefix"    : MsgFilter(logPrefixFilter),
                        "pretty_json"   : MsgFilter(prettyJson),
                        "fix_new_line"  : MsgFilter(fixNewline),
-                       "add_topic"     : MsgFilter(addTopic),
+                       "show_topic"     : MsgFilter(showTopic),
```
miki725, line 21: nit indent of `:`

miki725 (approved):

PR #112, opened 2023-12-20
fix: only wrapping ENTRYPOINT, if present

miki725:

## Issue

images with custom ENTRYPOINT from base image would break such as default AWS lambda images

## Description

Previously we were wrapping either CMD/ENTRYPOINT however that is incorrect behavior as it can change semantics of the image if the base image has any custom ENTRYPOINT

For example official AWS lambda images have this problem and their ENTRYPOINT requires specific CMD shape
and therefore we cannot adjust CMD without guaranteeing container behavior does not change.

---

In addition PR includes:

* switches cloud metadata to be json object vs json string
* reporting all container/task/stats metadata for ecs vs only container metadata which is limiting. task for example includes more information about task such as task arn/version/etc
* adding minimal lambda metadata collection. not much we can do there as lambda doesnt have endpoint aside from a few env vars but using those we can get some basic data such as aws account id, role arn, etc
* extracting some lambda/ecs metadata to dedicated chalk keys such as `_AWS_REGION`
* fixes various places which were assuming `resolvePath` can actually resolve which is not true in lambda for example which resulted in segfaults. the reason is that lambda runs with user `993`, not default container user which and as that user does not have home path anything which resolves `~` blows up
* small other things such as adding exception message when plugin blows up which is very helpful in debugging
* added `priority` field to sink configs to allow to prioritize some sinks higher in the processing order
* adds sink retries as we occasionally were seeing read timeouts. as its network related seems like a good idea to add timeouts there for safety

## Testing

```
➜ make tests args="test_docker.py::test_base_image --logs --pdb"
```

## Requires/TODOs

- [x] https://github.com/crashappsec/nimutils/pull/37 (current chalk is pinned to transient commit from nimutils)
- [x] add config to allow to force to wrap image even when ENTRYPOINT is missing


viega (approved):
Okay. Discussed comments off-line, and have been addressed.

PR #116, opened 2023-12-27
fix: added _ACTION_ID in push template and moved to the top of all report template configs

### LINKED ISSUE #115: _ACTION_ID missing on push_default template
## Description

`_ACTION_ID` should be present in all reports, but was missing from the push_default report template

## Steps to Reproduce

Build an image and then run the push command under chalk and observe the output does not contain an `_ACTION_ID` key:
`chalk docker push ghcr.io/crashappsec/test`

## Other Information

Would be useful to force `_ACTION_ID` as a key on all reports, regardless of template

nettrino:
## Issue

Fixes #115

## Description

This PR adds `_ACTION_ID` to the default `push` template
## Testing

Issue a `push` command and see that `_ACTION_ID` is properly set

### DIFF
--- a/src/configs/base_report_templates.c4m
+++ b/src/configs/base_report_templates.c4m
@@ -1667,6 +1667,9 @@ report_template push_default {
   You can certainly collect a lot more data here, but we expect it
   should usually be redundant w/ the info you collected during a build.
 """
+  key._ACTION_ID.use                          = true
+  key._TIMESTAMP.use                          = true
+  key._DATETIME.use                           = true
   key._OPERATION.use                          = true
   ~key._CURRENT_HASH.use                      = true
   key._STORE_URI.use                          = true


src/config.nim
```diff
@@ -30,8 +30,15 @@ template getReportTemplate*(): ReportTemplate =
   let
     outconf  = chalkConfig.outputConfigs[getBaseCommandName()]
     tmplName = outconf.report_template
+    actionId = "_ACTION_ID"
+  let templateRef = chalkConfig.reportTemplates[tmplName]
```
nettrino, line 34: @viega not sure if this is the cleanest spot to do this or if I should just use `forceReportKeys` on `src/reporting.nim`

miki725: wasnt force keys meant for this use-case where some keys are forced to be always reported like metadata id?

tests/data/configs/profiles/minimal_profile.c4m
```diff
@@ -1,9 +1,10 @@
 report_template minimal_report_template {
+    key._ACTION_ID.use    = true
     key._OPERATION.use    = true
     key.MAGIC.use         = true
     key._CHALKS.use      = true
```
miki725, line 5: nit. align `= true` with the others

miki725 (approved):
👍

miki725:
merging so that I can rebase the other PR to start sending those fields

Issue #95, opened 2023-11-10
add test for sast integration


indecisivedragon:
from https://github.com/crashappsec/chalk/pull/94, semgrep output on some build or insert should be validated

PR #129, opened 2024-01-11
test: add sast test with semgrep

indecisivedragon:

## Issue

closes https://github.com/crashappsec/chalk/issues/95

## Description

add test for semgrep


tests/data/python/sample_1/helloworld.py
```diff
@@ -2,6 +2,10 @@
 # dummy comment second line
 def main():
     print("Hello World")
+    # this is added for semgrep test
+    test_var = "aaa"
+    if test_var is "bbb":
```
miki725, line 7:
```suggestion
    if test_var == "bbb":
```

`is` compares singletons which strings are not guaranteed to be singleton

python internally caches small strings so in most cases its true but its def not true for long strings:

```python
>>> "a" is "a"
<stdin>:1: SyntaxWarning: "is" with a literal. Did you mean "=="?
True

>>> "a" * 2**20 is "a" * 2**20
False
```

tests/test_plugins.py
```diff
@@ -712,3 +713,120 @@ def test_syft_binary(copy_files: list[Path], chalk_copy: Chalk):
...
+    # check that sbom has been embedded into the artifact
+    chalk_mark = ChalkMark.from_binary(tmp_data_dir / "helloworld.py")
+    assert chalk_mark.contains(sast_data)
+
+    # logger.info("after insert")
+
+    # text = (tmp_data_dir / "helloworld.py").read_text()
+    # logger.info(text)
+
+    # time.sleep(1000)
```
miki725, line 832: forgot to cleanup?

indecisivedragon: yup, whoops

tests/data/python/sample_1/helloworld.py
indecisivedragon: we need this to be `is` for the semgrep rule to trigger correctly (`rule,yaml`)

viega (approved):

PR #137, opened 2024-01-17
refactor(no-api-login)!: remove the --no-api-login switch + assoc. tests

MyNameIsMeerkat:

## Issue

No issue

## Description

Remove the `--no-api-login` switch as it longer has relevance since the `chalk setup` flow changed and no longer requires an interactive OIDC authentication. 

Tests that included the switch have also been refactored.

### DIFF
--- a/src/configs/chalk.c42spec
+++ b/src/configs/chalk.c42spec
@@ -2755,15 +2755,6 @@ Chalk will also embed the keypairs internally, for future operations.
 """
 }
 
-  field api_login {
-    type:     bool
-    default:  true
-    shortdoc: "Use CrashOverride API"
-    doc: """
-Enable the use of the Crash Override API.
-"""
-  }
-
   field ignore_patterns {
     type:     list[string]
     default:  [".*/\\..*", ".*\\.txt", ".*\\.json"]
--- a/src/configs/getopts.c4m
+++ b/src/configs/getopts.c4m
@@ -661,11 +661,6 @@ with `.pub` as the file extension.
 """
   }
 
-  flag_yn api_login {
-    field_to_set: "api_login"
-    doc: """Enable/disable the use of the Crash Override API for secret management."""
-  }
-
   command gen {
     shortdoc: "Force generation of new signing key"
     doc: """
--- a/tests/chalk/runner.py
+++ b/tests/chalk/runner.py
@@ -219,7 +219,6 @@ class Chalk:
         exec_command: Optional[str | Path] = None,
         as_parent: Optional[bool] = None,
         no_color: bool = False,
-        no_api_login: bool = False,
         params: Optional[list[str]] = None,
         expected_success: bool = True,
         expecting_report: bool = True,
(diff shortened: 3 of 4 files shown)


MyNameIsMeerkat:
> 1. This is technically a breaking change, right?

Technically yes you are right, if anyone used that flag and tried to use it in future then it would return an error of `Invalid flag` 

> 2. If yes to 1, the commit message type shouldn't be `refactor` as per the semantic commit message format, right? If we at least partially generate the release notes from the commit message format, that'll help ensure that this change appears in the user-facing changes.

I chose refactor as this was part of the large `chalk setup` refactor but was just missed from the original PR. Happy to change it to BREAKING CHANGE if needed. Release notes are not currently autogenerated.

> 3. Is it worth making chalk produce a specific error message for a user that tries this switch after its removal?

No, this chalk setup feature had zero real world use from users based on backend telemetry and the flag was only implemented to facilitate our own test scripts. We are waaaay pre-1.0 and do not support any versions prior to whatever the most current one is with multiple breaking changes expected over time. No need to add complexity to the code base around this.

miki725 (commented):
does this require changes to the autocomplete scripts?

MyNameIsMeerkat:
> does this require changes to the autocomplete scripts?

Not that I can see:

```
function _chalk_setup_either {
            COMPREPLY=($(compgen -W "--color --no-color --help --log-level --config-file --enable-report --disable-report --report-cache-file --time --no-time --use-embedded-config --use-external-config --no-use-external-config --show-config --no-show-config --use-report-cache --no-use-report-cache --debug --no-debug --skip-command-report --no-skip-command-report --symlink-behavior --store-password --no-store-password --key-file" -- ${_CHALK_CUR_WORD}))
}

function _chalk_setup_completions {
    case ${COMP_WORDS[${_CHALK_CUR_IX}]} in
        gen)
            _chalk_setup_either
            ;;
        load)
            _chalk_setup_either
            ;;
        *)
            COMPREPLY=($(compgen -W "--color --no-color --help --log-level --config-file --enable-report --disable-report --report-cache-file --time --no-time --use-embedded-config --use-external-config --no-use-external-config --show-config --no-show-config --use-report-cache --no-use-report-cache --debug --no-debug --skip-command-report --no-skip-command-report --symlink-behavior --store-password --key-file gen load" -- ${_CHALK_CUR_WORD}))
            ;;
        esac
}
```

MyNameIsMeerkat:
> With this PR, `--no-api-login` is still mentioned in the docs in this repo. Do we want to update the below too?
> 
> https://github.com/crashappsec/chalk/blob/43bf76b977f6ce197b5fb067584f981302791fb6/src/docs/core-secret-manager-api.md?plain=1#L54-L64

No, all of the internal chalk docs should get sourced from the upstream docs repo and the secrets manager docs have already been amended there.

The dangling docs here are a result of the unclear documentation process / flows, the source of truth should be the docs repo rather than updating the docs in multiple places.

This needs to be fixed in the grander scheme of things but this is not the PR to do that.

Issue #138, opened 2024-01-19
consider having a single source of truth for chalk version


ee7:
## Description

Currently, we have to update the chalk version in two places for every release:

https://github.com/crashappsec/chalk/blob/7a390c8822b69e2b931c0692cbc921b4026e484c/src/configs/base_keyspecs.c4m#L55
https://github.com/crashappsec/chalk/blob/7a390c8822b69e2b931c0692cbc921b4026e484c/chalk.nimble#L1

Historically, that's caused the versions to go out-of-sync: https://github.com/crashappsec/chalk/issues/28.

Happily, chalk has since received a test to ensure that doesn't happen again (https://github.com/crashappsec/chalk/pull/41), but it's still probably marginally worthwhile to define the version in one place.

We can do that by making the nimble file read `base_keyspecs.c4m`.

This isn't a high-priority user-facing thing, but it'd also help with making a development version of chalk write a distinguishable `CHALK_VERSION` from a release version of chalk, if that's something we deemed worthwhile.

## Result

A single source of truth for the chalk version, at the cost of a few extra lines and a `staticRead` in the nimble file.

## Dependencies

None.

## Subtickets

None.

Issue #28, opened 2023-10-03
chalk mark appears to contain incorrect `CHALK_VERSION`


ee7:
### Example

With a binary `foo`, run:

```shell
chalk insert foo
strings foo | grep CHALK_VERSION | jq -r '.CHALK_VERSION,.INJECTOR_COMMIT_ID'
```

with Chalk version 175f32f001d90de02c3efa10d19467f08e98a1a7.

### Expected behavior

```text
0.1.2
175f32f001d90de02c3efa10d19467f08e98a1a7
```

### Observed behavior

```text
0.1.1
175f32f001d90de02c3efa10d19467f08e98a1a7
````

### Discussion

Maybe I'm missing something? But the Chalk version is reported as `0.1.2` when running `chalk version`:

```console
$ chalk version | head -n3
┏┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┳┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┉┓
┋ Chalk version ┋ 0.1.2                                    ┋
┋ Commit ID     ┋ 175f32f001d90de02c3efa10d19467f08e98a1a7 ┋
```

and all the versions are all the same in e.g. the [Getting Started - Chalk Insertion](https://crashoverride.com/docs/chalk/getting-started#chalk-insertion) example.

`CHALK_VERSION` is also different in `chalk.log`:

```console
$  jq < chalk.log | grep 'VERSION' 
        "CHALK_VERSION": "0.1.1",
    "INJECTOR_VERSION": "0.1.2",
    "_OP_CHALKER_VERSION": "0.1.2",
```

viega:
Yes, thanks for finding this. We need to write a script that bumps the version everywhere in the sample place!

PR #140, opened 2024-01-19
refactor: set version in nimble file from base_keyspecs

ee7:
To-do:

- [x] Consider removing the validation (parsing into a string, not three ints) if it's sufficiently validated elsewhere. The overkill approach would be a user-defined matcher.
- [x] Get CI to pass, updating these (and anything else that's necessary):

https://github.com/crashappsec/chalk/blob/f496958f9e3078f7f52b5e84bfdce5847dee6f26/Makefile#L24

https://github.com/crashappsec/chalk/blob/f496958f9e3078f7f52b5e84bfdce5847dee6f26/src/chalk_common.nim#L360

## Issue

#138

## Description

Before this PR, we had to update the chalk version in two places:

https://github.com/crashappsec/chalk/blob/991a87f989965994bede3539e793b859ff1ced47/src/configs/base_keyspecs.c4m#L55
https://github.com/crashappsec/chalk/blob/991a87f989965994bede3539e793b859ff1ced47/chalk.nimble#L1

Historically, this duplication has caused the versions to go out of sync such that a [chalk mark contained an incorrect `CHALK_VERSION`][1]. The tests in chalk now guard against that (2f1b55cd70bf), but it's still better to define the version in one place.

Make the keyspecs version the single source of truth, setting the version in the nimble file from there.

This will also help with making a development version of chalk have a distinguishable version string from a release version of chalk, if that's something we deem worthwhile.

Also, refactor to remove the macro that was previously used in setting the version.

Closes: #138

[1]: https://github.com/crashappsec/chalk/issues/28

## Testing

This PR updates the `chalk version` test:

https://github.com/crashappsec/chalk/blob/d1c14ff3ee1a4e5c09d4ea0920600833f6c1e719/tests/test_command.py#L155-L167

~but I think I'll add a test there to check the `make version` output is the same.~


viega (dismissed):
Much better, thank you.

miki725 (approved):

Issue #177, opened 2024-01-29
Readme.md link for CLA is broken


nettrino:
## Description

The CLA link in the Readme yields a 404

miki725:
duplicate of https://github.com/crashappsec/marketing-site/issues/201

PR #179, opened 2024-01-30
Update contributing link

nettrino:
## Issue

Fixes https://github.com/crashappsec/chalk/issues/177

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ We are constantly learning about emerging use cases for Chalk, and are always in
 
 ## Making contributions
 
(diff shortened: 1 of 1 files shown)


miki725 (changes requested):
lets fix in the docs itself. that link was used across the org so chalk is not the only place where the link is broken now

miki725:
the deploy fo the docs site has failed which is why its not reflected yet

miki725:
```
➜ curl https://crashoverride.com/docs/other/contributing -Is | grep -i location
location: /docs/other/overview
```

link is working now

Issue #181, opened 2024-01-31
Nim imports could be clearer and more robust


ee7:
## Description

Nim imports in this repo tend to look like this:

https://github.com/crashappsec/chalk/blob/ae84275da429dcb3b5afcc5953318045340f96ee/src/util.nim#L11-L12

We can make this more readable and more robust.

## Result

I suggest using this style:

```Nim
import std/[tempfiles, osproc, posix, monotimes, parseutils]
import pkg/[nimutils/managedtmp]
import "."/[config, subscan]
```

## Dependencies

None.

PR #182, opened 2024-01-31
refactor: improve robustness and clarity of Nim imports

ee7:
- [x] Double-check the diff myself
- [x] Set as ready to review

## Issue

https://github.com/crashappsec/chalk/issues/181

## Description

There was no consistent style for Nim imports in this repo, but imports tended to be all grouped together. For example:

```nim
import std/tempfiles, osproc, posix, config, subscan, nimutils/managedtmp,
       std/monotimes, parseutils
```

I believe that the best practice is to write that like this:

```nim
import std/[tempfiles, osproc, posix, monotimes, parseutils]
import pkg/[nimutils/managedtmp]
import "."/[config, subscan]
```

Apply that style for imports across the codebase. To be specific:

1. First import any Nim standard library packages, and use the `std/` prefix.
2. Then import any external Nim package, and use the `pkg/` prefix.
3. Then import any repo-local module, using explicit paths, and use a common path prefix where possible.
4. Import only the necessary modules.

This improves:

- Maintainability and readability: it's faster for a chalk developer to understand the dependencies for a module. For example, it is helpful to immediately see which repo-local modules are imported.
- Robustness: if two modules have the same name, the one we're importing is explicit. For an example problem that can otherwise occur, see https://github.com/nim-lang/Nim/commit/2b1a671f1cbe4d3443537435ef0f0cddf928abbf.
- Consistency: newer Nim standard library modules can only be imported with the `std/` prefix, but there's no good reason to separate those from the other standard library imports at the import site.

This commit is also in line with the [Nim Manual's recommendations][nim-manual]:

>There are two pseudo directories:
>
>1. `std`: The `std` pseudo directory is the abstract location of Nim's standard library. For example, the syntax `import std / strutils` is used to unambiguously refer to the standard library's `strutils` module.
>2. `pkg`: The `pkg` pseudo directory is used to unambiguously refer to a Nimble package. However, for technical details that lie outside the scope of this document, its semantics are: *Use the search path to look for module name but ignore the standard library locations*. In other words, it is the opposite of `std`.
>
> It is recommended and preferred but not currently enforced that all stdlib module imports include the std/ "pseudo directory" as part of the import name.

See also these recent commits in the Nim compiler repo, which now also imports every stdlib module using the `std/` prefix:

- https://github.com/nim-lang/Nim/commit/e17237ce9dbf5410623e9d510217e7817bf4fd89
- https://github.com/nim-lang/Nim/commit/4d11d0619d11d8aecdaff998981efcd054de9760

Closes: https://github.com/crashappsec/chalk/issues/181

[nim-manual]: https://github.com/nim-lang/Nim/blob/v2.0.0/doc/manual.md#pseudo-importinclude-paths

## Testing

`git checkout` this PR's branch, then in the root directory of this repo, grep for `import` in Nim files:

```shell
git grep --break --heading --context 1 'import ' -- '*.nim' 
```

and check for consistency.

Sadly, I am not aware of an existing robust tool to lint Nim code for this style in CI.

Note that this PR is potentially a semantic change: it changes the order of imports, and import ordering can be meaningful in Nim.

For that reason, and to minimize the diff size, I deliberately did not alphabetize imports in this PR. I consider that less important, but still worthwhile, so I'll suggest doing that later.


viega:
While I won't stand in the way of doing this, I'm philosophically not aligned to the official recommendations:

1) In a world where language servers are ubiquitous, it's very easy to figure out how code interrelates across modules. People don't need to take on the extra mental burden just because Nim doesn't have a working LSP :)
2) Nim is particularly bad at requiring imports for functionality that should just be omnipresent, and uses some names that are not the most obvious.
3) In some cases, due to the fact that Nim can't deal well with cross-module dependencies, the order of imports turns out to matter (though I've mostly alleviated that once I figured out I can basically force C prototypes to solve the problem).
4) Generally, programmers have enough mental burden. They shouldn't have to worry about names of modules for really common stuff (and Nim has some module names I still can't remember after more than a year).

Note that we do have *some* code that runs at the top level, and thus the order of imports becomes semantically important.

Generally, I would prefer to stuff most of them as far down the dependency tree as possible and re-export, so that people don't have to worry about it. But often, in the moment, it's hard to know how things are going to evolve, and again, Nim is pretty intolerant here due to its poor cross-module dependency handling.

viega (dismissed):
It looks particularly odd to me to add quotes for imports. I get that it's available for cases where the file name isn't a valid identifier, but it's surprising / unfamiliar / irregular.

I'm generally fine with extra typing that improves clarity, but that is even more baffling than putting "std/" in front of things that a) don't need it and b) shouldn't have to have it even if the language did require it :)

viega:
> It's true that Nim has made deliberate design decisions to not optimize for "readability when the code is printed on paper", but I think it's still helpful to explicitly state what kind of module is imported: standard library, external dependency, or repo-local.

Sure, though again, most of the standard library type stuff in my view should be omnipresent and hidden.

> I might argue that the continued lack of solid Nim tooling is a reason to make things as clear as possible in the meantime :)

More people are going to look at the code than try to dig into it, and I think overwhelming numbers of imports tends to confuse and comes across as daunting, but that's a personal opinion for sure.

> > Nim is particularly bad at requiring imports for functionality that should just be omnipresent, and uses some names that are not the most obvious.
> 
> I guess it just depends on personal taste and background whether more built-ins makes things easier or harder? I'm OK with the amount that Nim has, but I'm used to it, and people have definitely argued for less.

I think in any world, basic operations around built-in data types should not need to be imported. That includes stuff across strutils, sequtils, unicode, etc. I also don't like having to import syntax, because you shouldn't have to know to do that (e.g., sugar). 

> I think the consensus was that the system module was big already (and it was bigger in the past), and that's particularly painful when we don't have incremental compilation yet. I imagine that compiling `system.nim` over and over again is a significant fraction of the world's total Nim compilation time.

Compile times aren't great, but they're not too horrendous either. But I think that IC would have a huge practical impact in many ways, and if that were really a barrier for them adding more to the stdlib (which it doesn't seem to be), then I'd wonder even more why they keep procrastinating :)

> This PR deliberately preserves the existing order of imports within an import kind. Are you aware of anything that could break because e.g. a nimutils module is now imported before a repo-local module, rather than the other way around?

No, there should be no dependencies in that direction.


> I agree completely. But I'm making this PR because I find it to meaningfully decrease the cognitive load when reading the code, rather than increase it :)
> 
> I think it also makes the codebase more accessible to people who are new to the codebase, or less experienced in Nim (and therefore haven't memorised the names of every standard library module). I think it's helpful to immediately know that e.g. `unicode` is a standard library module, and `collect` isn't, without having to use editor features.
> 
> Do you find that the changes in this PR decrease readability for you?

My view is that hiding most of the imports is the right way to go, especially the Nim ones, possibly not omnipresent  dependencies like nimutils, though much of that is 'stuff that I don't like how it was done in the std libraries'.

But I don't have a problem w/ organizing them better; I have gone through phases there; often the end up ordered like that until something happens, like a naming conflict, or refactoring, then I never remember to go re-organize when the dust settles.

But it's enough of a pain, it's another reason why I prefer to export more.

> > It looks particularly odd to me to add quotes for imports. 

> I don't love this either. But yeah, it's required because this doesn't work:

Yeah, then I personally just would go w the list of individual items, instead of the grouping syntax. To me that's just a poor shortcut for requiring too many imports in the first place :)  But I will also live with it.  


> Consider the below situation:
> 
> 1. I write some project, and write `import foo` to import a local module or a Nimble package
> 2. Everything works fine
> 3. I upgrade Nim, which contains a new standard library module `foo`

Yeah, Nim needs to be smarter about namespaces. It's right to prefer the local name for local code, but the system should not have access to the local namespace at all when importing, it should only have access to the system namespaces. When the user wants to shadow, but still access the system namespace they should then be able to be explicit. 

I'd rather avoid 'std' everywhere and be more aware, because I think it looks like a big wart when people look at the code, and I don't want people to look at it and go, "that's gross" :) But that's my only real issue.

Note that I did approve the PR in the first place; I'm happy to accept everything, just grousing and would ultimately prefer the common stuff to all get hidden.

miki725 (approved):
👍 amazing!

PR #184, opened 2024-01-31
fix: chalk setup does not save existing keys

miki725:

## Issue

When `chalk.pub`/`chalk.key` and `CHALK_PASSWORD` is already provided, chalk would not save them into chalkmark. Chalkmark was only updated when new key material was being generated. As such it was not possible to reuse existing keys aside from sharing binary itself.

## Testing

```
➜ make tests args="test_command.py::test_setup_existing_keys --logs --pdb"
```

### DIFF
--- a/src/configs/chalk.c42spec
+++ b/src/configs/chalk.c42spec
@@ -2676,7 +2676,7 @@ validate in the transparency log.
 
   field use_signing_key_backup_service {
     type:    bool
-    default: true
+    default: false
     shortdoc: "Use signing key backup service to save & retrieve keys"
     doc: """
 Any signing keys generated or imported are encrypted by a randomly
--- a/configs/connect.c4m
+++ b/configs/connect.c4m
@@ -31,6 +31,7 @@ custom_report crashoverride {
   ~use_when:        ["insert", "build", "exec"]
 }
 
+use_signing_key_backup_service       = true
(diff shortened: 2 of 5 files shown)


MyNameIsMeerkat (approved):
Changes look like they make sense to me, this compiles and tests pass for me locally as well.

LGTM 👍

src/configs/chalk.c42spec
```diff
@@ -2676,7 +2676,7 @@ validate in the transparency log.
 
   field use_signing_key_backup_service {
     type:    bool
-    default: true
+    default: false
```
miki725, line 2679: changing this default to `false` as otherwise OSS chalk shows errors on `chalk setup`:

```
error: auth_config.crashoverride is referenced but its missing in the config
error: Could not retrieve Chalk Data API token from configuration profile. Unable to use Signing Key Backup Service.
error: Could not store password. Either try again later, or use the below password with the CHALK_PASSWORD environment variable. We attempt to store as long as use_signing_key_backup_service is true.
```

to enable it for the console integration, am enabling it in `connect.c4m` as thats what defines `auth_config.crashoverride`

https://github.com/crashappsec/chalk/blob/de311771915d36e712baa36a23e47cd9618a93ef/configs/connect.c4m#L34

cc @viega

viega (approved):

Issue #186, opened 2024-02-06
401 from chalk api segfaults


miki725:
## Description

when working on openid at some point was putting invalid json tokens into the config and saw this error:

```
trace: Signing Key Backup Service status code: 401 Unauthorized

trace: Signing Key Backup Service response: Unauthorized data read

parsejson.nim(518)       raiseParseErr
```

## Impact

as the container was wrapped with chalk, segfault exited the whole container without any easy resolution path. we should figure out what is causing the issue

## Steps to Reproduce

load invalid jwt and run wrapped container image

miki725:
also why is signing service is required during exec? if we need to validate any signatures, that should only use public key so password for the private key should not be required

cc @MyNameIsMeerkat am I missing something?

miki725:
after debugging with @MyNameIsMeerkat we noticed a couple of issues here:

* when providing cosign key material, chalk sends the pass to the backup service which is incorrect. when all the data is present, why send it to backup?
* as we generate random jwts (either in UI or via openid connect), multiple jwts can be associated with the same key id which backup service will need to handle
* backup service should only be invoked when needing the password to decrypt the private key. otherwise commands like `exec` should not need to invoke backup service at all
* getting 401 from the service causes segfault so we are prolly not catching some condition correctly somewhere in `attestation.nim`

PR #189, opened 2024-02-06
fix: handling non-json responses from signing backup service

miki725:

## Issue

https://github.com/crashappsec/chalk/issues/186

## Description

not in all cases service returns valid json for non-200 responses and so we should not parse it assuming it has valid json which caused segfault

## Testing

point signing service to dummy endpoint which doesnt return json back


MyNameIsMeerkat (approved):
Yup that logic change makes sense to me 👍

PR #193, opened 2024-02-08
build: using scratch for base image of public chalk image

miki725:

## Issue

make it easy to copy chalk into other containers

## Description

As chalk is a static binary now there is a no need to release docker image based on any distro as we can just release the raw binary image. This image is not meant to be used directly but is instead meant to be used in other docker builds such as to easily provision docker build images used in CI systems.

Once merged Ill push image to registries

## Testing

```
docker build -t chalk .
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,7 +44,6 @@ WORKDIR /chalk
 # vs COPY . /chalk/
 # as repo has other tools and copying only necessary files
 # optimizes docker build cache
-COPY --chmod=755 bin/devmode /chalk/bin/
 COPY config.nims /chalk/
 COPY ./src/ /chalk/src/
 # for chalk commit id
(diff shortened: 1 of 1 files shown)


miki725:
that image is not meant to be used directly. its meant to be used as:

```dockerfile
FROM ...
COPY --from=chalk /chalk /usr/bin/local/chalk
```

dont imagine people will ever want to directly base their images on top chalk as its a util and will require other misc utils at runtime during various commands like `git`/`docker`/etc

for example to add chalk to existing docker image will be something like:

```dockerfile
FROM docker
COPY --from=chalk /chalk /bin/chalk
```

as such dont think we need to worry about certs in base chalk image. although that should be addressed by embedding certs in chalk which hopefully will happen soon

for now lets keep it simple with scratch-only image. if users will ask for additional images, we can of-course provide those at a later time

miki725:
> Sounds right to me. But we didn't have some special reason for wanting the previous chalk:alpine and chalk:ubuntu images?

no it was just a placeholder as initially we wanted to release chalk as a docker image so you can do:

```
docker run --rm -it chalk ....
```

then then we formalized the release pipeline which publishes binaries so there was no need for any of the docker stuff hence that was never actually used

---

> Sure. Although with this installation method, I guess it's more important to document any such requirements.

yes we will need to document all available install methods which will be:

* download binary directly
* use setup.sh with curl pipe
* docker copy
* AUR packages

---

> This was discussed a bit on a call, but I didn't see a ticket for it. Do we have one somewhere?

not sure. dont see one either but maybe its in another repo? It was a pretty high priority task to work on in near future

PR #195, opened 2024-02-08
build: adding -dev support to chalk version

miki725:

## Issue

dev builds when chalk is not on a release results in showing version as if its for a release

## Description

this will allow to bump version to `*-dev` immediately after the release to denote chalk is for a dev version

this is especially useful for publishing chalk to other platforms such as to docker hub/etc

## Testing

as nimble does not support suffixes, its ignored there:

```
➜ nimble version | tail -n1
0.3.3

➜ make version
0.3.3-dev

➜ ./chalk version 2> /dev/null | grep -i version
 ┊ Chalk Version  ┊ 0.3.3-dev                                   ┊
```

### DIFF
--- a/src/semver.nim
+++ b/src/semver.nim
@@ -11,40 +11,53 @@ import std/strutils
 # it only handles basic dot-separated version format
 
 type Version* = ref object
-  major: int
-  minor: int
-  patch: int
-  name:  string
+  major:  int
+  minor:  int
+  patch:  int
+  suffix: string
+  name:   string
 
 proc parseVersion*(version: string): Version =
   var
-    major = 0
-    minor = 0
-    patch = 0
+    major  = 0
+    minor  = 0
+    patch  = 0
+    suffix = ""
   let
-    name  = version.strip(chars={'v', ',', '.'})
-    parts = name.split('.')
+    name     = version.strip(chars={'V', 'v'}, trailing=false).strip(chars={',', '.', '-', '+'})
+    sections = name.split({'-', '+'}, maxsplit=1)
+    parts    = sections[0].split('.')
   case len(parts):
     of 1:
-      major = parseInt(parts[0])
+      major  = parseInt(parts[0])
     of 2:
-      major = parseInt(parts[0])
-      minor = parseInt(parts[1])
+      major  = parseInt(parts[0])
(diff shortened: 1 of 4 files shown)


src/semver.nim
```diff
@@ -11,40 +11,53 @@ import std/strutils
...
-    minor = 0
-    patch = 0
+    major  = 0
+    minor  = 0
+    patch  = 0
+    suffix = ""
   let
-    name  = version.strip(chars={'v', ',', '.'})
-    parts = name.split('.')
+    name     = version.strip(chars={'v'}, trailing=false).strip(chars={',', '.', '-', '+'})
```
MyNameIsMeerkat, line 27: Maybe also want to strip out a capitalised 'V' prefix ?

miki725: sure why not

miki725: done

src/semver.nim
```diff
@@ -11,40 +11,53 @@ import std/strutils
...
 # version parts tuple used for comparison
-proc parts(self: Version): (int, int, int) =
-  return (self.major, self.minor, self.patch)
+proc parts(self: Version): (int, int, int, string) =
+  # TODO how to compare suffix?
+  # for now treating any suffix as less than no suffix
+  # assumping any suffix is for pre-releases which is not ideal
+  # correct but it is fine for chalk versions
+  # this handles things like 1-dev < 1.0
+  # no suffix is normalized to highest ascii char code \u7f
```
MyNameIsMeerkat, line 57: Is there any risk to defaulting to a suffix of "\u7f" rather than an empty string "" for when there isn't a suffix present? Will this change break any calling code that would be surprised to get a "\uf7" value as a suffix?

MyNameIsMeerkat (approved):
Couple of small comments but certainly nothing blocking I could see 👍

src/semver.nim
miki725: to clarify, if there is no suffix, the `\u7f` is added to both versions being compared so they will be equal. the difference comes into play only when one of the versions doesnt have the suffix

MyNameIsMeerkat: Thanks for the additional clarity @miki725 and agree with you both that this seems more than good enough for now 👍

PR #200, opened 2024-02-09
fix: rename JSon to Json or JSON

ee7:

## Issue

n/a.

## Description

There was some inconsistency in JSON-related identifiers in the codebase. For example, on the middle left of:

https://github.com/crashappsec/chalk/blob/95691512614e7e305c7f477cbc7c5608b2a7f32a/src/chalkjson.nim#L223-L225

Nim is partially style insensitive, but it's best to make identifiers consistent to best support e.g. searching for usages with `grep`.

The [standard capitalization](https://www.json.org/json-en.html) is JSON (JavaScript Object Notation), not JSon. So let's rename:

- `JSonNode` to `JsonNode`, which is [the upstream style][1]
- `CJSonError` to `CJsonError`
- `ChalkJSonNode` to `ChalkJsonNode`
- `JSon` to `JSON` in comments and exception messages

Since we're using semantic commit messages: this is *technically* a user-facing fix, because it affects an exception message.

[1]: https://github.com/nim-lang/Nim/blob/v2.0.0/lib/pure/json.nim#L188

## Testing

```console
$ git rev-parse --short HEAD
533cbf9
$ git grep 'JSon' || echo 'none'
none
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,7 +44,6 @@ WORKDIR /chalk
 # vs COPY . /chalk/
 # as repo has other tools and copying only necessary files
 # optimizes docker build cache
-COPY --chmod=755 bin/devmode /chalk/bin/
 COPY config.nims /chalk/
 COPY ./src/ /chalk/src/
 # for chalk commit id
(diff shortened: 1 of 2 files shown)


miki725:
this confused me once so thanks 👍

PR #222, opened 2024-02-27
fix: honor $CHALK_COMPONENT_CACHE in confload

miki725:

## Issue

containers were failing to startup as `connect.c4m` was changed after the release in the chalkdust bucket

## Description

when loading components on chalk startup, if a component was from a URL and had any parameters, it would be re-fetched from the URL again which is not desired as the URL content can change hence causing a failure.

this was caused by loading parameters before loading component source which would initialize the component and if missing would re-fetch it again from URL. By loading source first it removes that possibility.

## Testing

load components from URL and then ensure they dont change on load

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 - Fixes a segfault when using secrets backup service
   during `chalk setup`
   [#220](https://github.com/crashappsec/chalk/pull/220)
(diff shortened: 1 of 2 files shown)


miki725:
a lot of the component loading machine will change with v2. the exported proc I think is just by convention. its not used anywhere else in the chalk codebase

Issue #226, opened 2024-02-29
CI: should run unit tests


ee7:
### Proposal

Chalk is a binary, not a library, so the most important thing to test is the binary itself. We currently use pytest for that.

However, unit tests can be useful. I'd like to set up testing of Nim procedures in CI. 

### Dependencies

None.

### Subtickets

None.

PR #227, opened 2024-02-29
ci: add testament-based unit testing

ee7:
### Description

Move the only unit tests we had to a dedicated directory, and set up running unit tests in CI. The point is to make it easy to add more unit tests in the future. For example, I wanted to add some when I was writing https://github.com/crashappsec/chalk/commit/920bdf55711549dba32f4babc1e2738cf795ab28 to illustrate some parsing of git objects.

To run unit tests, the user should run `nimble test`. This will use `testament` to test all Nim files beginning with `t` in the `tests-nim` directory.

~The macro uses a new `getSortedPaths` proc, which we can consider moving to nimutils later.~ I intend to refactor [`staticListFiles`](https://github.com/crashappsec/nimutils/blob/a7ec24b27453342157cefe947e5730af5f5c05af/nimutils/filetable.nim#L19-L35) at some point to work something like this. We can avoid parsing `ls`.

For more details, see the individual commit messages.

Closes: https://github.com/crashappsec/chalk/issues/226

### Testing

- Look at the CI log, and verify that the tests ran.
- Add `doAssert false` to `tests-nim/test_semver.nim` locally, run `nimble test`, and verify that the tests error.

---

To-do:

- [ ] Consider being consistent in calling this "nim tests" or "unit tests", and changing file/dir names accordingly. The latter might be better for now, but maybe we could have some integration tests in Nim in the future.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 .TODO
 /bin
-/bin/chalk
 /chalk
 /src/getlibpath
 /deps


.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,41 @@
+name: unit-tests
+
+on:
+  push:
+    branches:
+      - main
```
miki725, line 6: please copy all the other qualifiers so that we dont run tests on for example on README changes/etc

https://github.com/crashappsec/chalk/blob/b583d27c6648b6860149b3b3550a29185c9015ed/.github/workflows/tests.yml#L7-L13

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,41 @@
...
+      fail-fast: false
+      matrix:
+        include:
+          - os: linux
+            runs-on: ubuntu-22.04
+            arch: x86_64
+
+    name: "${{ matrix.os }}-${{ matrix.arch }}"
+    runs-on: ${{ matrix.runs-on }}
+
```
miki725, line 25: please add `concurrency` which will cancel previous commit runs if new commits are pushed on the pr

https://github.com/crashappsec/chalk/blob/b583d27c6648b6860149b3b3550a29185c9015ed/.github/workflows/tests.yml#L53-L58

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,41 @@
...
+        with:
+          version: "binary:2.0.0"
+        env:
+          GITHUB_TOKEN: ${{ github.token }} # Avoid rate limiting.
+
+      - name: Install Nimble dependencies
+        run: nimble --accept install --depsOnly
+
+      - name: Run unit tests
+        run: nim r tests-nim/all_nim_tests.nim
```
miki725, line 41: so far we normalize everything on docker image which has all deps pre-installed and is fetched from the github container registry as cache. lets use a consistent way to run nim things in all of CI

also we use `make` for triggering how tests are executed so lets add `make unit-tests` or similar target which will then use all the makefiles machinery to run tests as we currently do for rest of the tests

miki725 (changes requested):
overall idea is good to run unit tests in CI

lets update tests/README.md as well documenting how to run tests

tests-nim/all_nim_tests.nim
miki725, line 1: there is no unit test runner in nim or a framework we can use for tests? this seems wild to manually collect all test files to run tests with

viega:
Agreed, and while I'd love to see automated unit testing, I'm wondering if
it is worth a significant amount of effort right now compared to all the
other things we have in the backlog? The unit tests that have been produced
have mainly been when first writing a module, and few of them have been
kept current as modules evolved, generally because they only covered basic
functionality that integration tests also manage.

.github/workflows/unit-tests.yml
miki725: path ignore is not equivalnt to path opt-in.

for example this will trigger tests when updating `.clabot` file which is not related to tests

please switch to use existing pattern for consistency

.github/workflows/unit-tests.yml
miki725: I feel like its still wasting some minutes somewhere but fair enough. but I strongly feel it should be consistent across the repo so if we dont want that behavior, lets match it everywhere

.github/workflows/unit-tests.yml
miki725: as it adds it here, lets keep it consistent with existing patterns in the repo. currently:

* we use makefile for triggering tests in CI/etc as it can easily orchestrate prepping additional files in the repo.
* use docker for all dep orchestration. this allows to exactly replicate any CI setup locally with ease. also it makes it easy to manage/update all deps without changing system things. of course if people would like to run things on system they can but its more of an advanced mode and we cannot guarantee everything will work whereas our docker image we can. both @indecisivedragon and me exclusively use 🐳 for compiling chalk/running tests so lets follow the same here. makefile can easily call nimble `test` task if that simpler and you can use existing `chalk` compose service to leverage all pre-installed deps in there already - `docker compose run --rm chalk nimble test`.

tests-nim/all_nim_tests.nim
miki725: testament seems find

tests-nim/README.md
miki725, line 1: note a fan of `tests-nim` directory. makes top level confusing

python tests are not married to being top-level in `tests` folder. if nim wants all tests be in `tests` folder my proposal is to:

* merge all tests to be in `tests`
* `tests/unit` will have nim unit tests
* `tests/pytest` or `tests/integration` will be pytest stuff

this way its more intuitive

miki725 (changes requested):
lets use 🐳 across all of CI and not sure about folder structure

Makefile
```diff
@@ -138,6 +138,10 @@ tests: docker-setup
 tests: $(BINARY) # note this will rebuild chalk if necessary
 	docker compose run --rm tests $(make_args) $(args)
 
+.PHONY: unit-tests
+unit-tests:
+	docker compose run --rm chalk nimble test args='$(args)'
```
miki725, line 143:
```suggestion
	$(DOCKER) nimble test args='$(args)'
```

tests/functional/data/dockerfiles/valid/base/entrypoint.sh
miki725, line 1: please revert changes here

tests/functional/data/dockerfiles/valid/base/Dockerfile
miki725, line 1: same here. please revert

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,49 @@
...
+      - main
+    paths-ignore:
+      - '**.md'
+      - 'server/**'
+      - 'tests/**'
+  pull_request:
+    paths-ignore:
+      - '**.md'
+      - 'server/**'
+      - 'tests/**'
```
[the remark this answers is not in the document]
miki725: my vote to just match things in this PR as its changing paths anyhow

src/semver.nim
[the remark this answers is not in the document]
miki725: all fd cache tests require some files to exist so might need to `touch` some files before actually running those tests but they do work when files exist, at least they used to when it got merged as I was using them for testing various flows

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,49 @@
...
+        with:
+          version: "binary:2.0.0"
+        env:
+          GITHUB_TOKEN: ${{ github.token }} # Avoid rate limiting.
+
+      - name: Install Nimble dependencies
+        run: nimble --accept install --depsOnly
+
+      - name: Run unit tests
+        run: make unit-tests
```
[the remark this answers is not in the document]
miki725: copy the steps and permissions from integration tests. it needs to auth to github registry to pull the cached image which will already have all deps pre-installed

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,77 @@
...
+
+      - name: Install Nim
+        uses: iffy/install-nim@0ed218de08059d18507144e6aee70e6bad0bc148 # v5.0.3
+        with:
+          version: "binary:2.0.0"
+        env:
+          GITHUB_TOKEN: ${{ github.token }} # Avoid rate limiting.
+
+      - name: Install Nimble dependencies
+        run: nimble --accept install --depsOnly
```
miki725, line 47: why do we need this? tests will run in 🐳

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,77 @@
...
+
+      # for tests to checkout dummy private repo
+      - name: Get GitHub Token
+        id: org-token
+        uses: crashappsec/action-github-app-token@main
+        with:
+          app_id: ${{ secrets.CHALK_GITHUB_APP_ID }}
+          private_key: ${{ secrets.CHALK_GITHUB_APP_PRIVATE_KEY }}
+
```
miki725, line 56: this is only used in integration tests to checkout dummy repos as noted in the comment. we should def not need it for unit tests

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,77 @@
...
+          username: ${{ github.repository_owner }}
+          password: ${{ github.token }}
+
+      - name: Bake images
+        run: |
+          docker buildx bake chalk --load
+
+      - name: Run unit tests
+        env:
+          GITHUB_TOKEN: ${{ steps.org-token.outputs.token }}
```
miki725, line 75: why do we need to pass org token here? in integration tests this is used for git context testing which clones private repo in the org. unit tests should not need it

.gitignore
```diff
@@ -1,8 +1,11 @@
 .TODO
+/bin
 /bin/chalk
```
miki725, line 3: if we ignore `/bin`, prolly no need to ignore this anymore?

src/semver.nim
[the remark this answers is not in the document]
miki725: there is review overhead/etc for additional PRs so my preference is bite the bullet once and move one. I can help with that once Im in a good checkpoint on my docker refactor branch

tests/functional/data/dockerfiles/valid/base/Dockerfile
miki725: this is still a 🛑 from me. not sure why it got marked as outdated. the diff is still there

image

tests/functional/data/dockerfiles/valid/base/entrypoint.sh
miki725: same here. 🛑 from me

image

miki725 (changes requested):

tests-nim/README.md
miki725: Yep. Perfect

miki725 (approved):

Issue #224, opened 2024-02-28
Support a NOP run mode beyond virtual chalking


nettrino:
## Description

There are cases where for whatever reason a chalk bug could be affecting the build or execution of an image / binary etc. While we can do our best to safeguard against errors (see #188 and similar issues), we should offer a well-tested, robust `nop` mode of execution, where chalk only passes execution to the underlying command (e.g., if doing `chalk docker build <args>` to just do `docker build <args>`). This will be different than standard chalk operations like virtual chalking, in the sense that there will be 

* no con4m related initialization or code executing
* no requirement for setup/env etc. 
* no chalk flag like `trace` etc. respected

This will offer an easy escape path for users in case of critical bugs, as well as a quick debug flag to pinpoint whether an issue is related to chalk's side effects vs not.

## Result

Define an environment variable (e.g., `__CHALK_NOP`), which, when set, will result in a completely separate path of execution, which is dependency free, and only calls execve with the underlying arguments passed to chalk.

viega:
Well, before too long, the con4m code will always be pre-executed, except
for the command line flag parsing, the user's config and a few other very
minor things.

We could potentially skip user config loading there, but not the command
line parsing.

And, if people want to disable across an entire machine, it seems like we'd
WANT to respect that in the config file?  The config file can (e.g.,
through a parameterized module) implement the environment variable
semantics, but also let people just hardcode it off without worrying about
the environment if necessary.

That to me seems like the best overall approach; using the info can happen
before any other docker-related stuff happens (where bugs that need to be
worked around are likely to be).

viega:
100%. 99% of the con4m code is data structure setup that can all run statically, and then be frozen in memory until needed basically. 

My point is that it soon will not be bypassing much actual work to bypass the con4m execution, and it's unclear there are benefits from doing so, since that's currently responsible for the user interface for options like this... 

It's very easy to avoid *all* of the work Chalk would do though, so I see it as very reasonable to add a `bypass` mode, I just am not convinced that it should bypass the core con4m load.

I think for skipping the user-defined con4m config load, what's in place already would also suffice. So for me, `bypass mode` would be about bypassing all chalk processing.

miki725:
right, with con4m v2 it will remove a lot of the loading complexity.

for now I guess we can def add support for some env var to bypass things like loading custom configs, sinks/etc. basically with that env var chalk should not do anything error-prone such as doing network calls/etc

once we have v2 I think we can make the bypass a lot more comprehensive

[...]

PR #228, opened 2024-03-01
feat: allow bypassing chalk via CHALK_BYPASS environment variable

ee7:
## Description

@nettrino and I had a good discussion earlier today about bypassing chalk (issue https://github.com/crashappsec/chalk/issues/224). Here's a sketch of a possible implementation.

This is the "avoid as much chalk as possible" flavor, optimizing for a completely independent, minimal code path.

It allows bypassing the exec command:

```console
$ CHALK_BYPASS=1 chalk exec echo hi
chalk: the CHALK_BYPASS environment variable is set.
Running the following command without chalk:
  echo hi
hi
```

and bypassing the docker command:

```console
$ CHALK_BYPASS=1 chalk docker --version
chalk: the CHALK_BYPASS environment variable is set.
Running the following command without chalk:
  docker --version
Docker version 25.0.3, build 4debf411d1
```

and errors when the command is not docker or exec:

```console
$ CHALK_BYPASS=1 chalk version
error: CHALK_BYPASS was set, but the chalk command is not 'docker' or 'exec'. Quitting.
```

Refs: #224

## Testing

Run commands like illustrated above.

## To-do

If we wanted to continue in this direction:

- [ ] Tests
- [ ] Docs
- [ ] Consider handling `chalk` being named `docker`

### DIFF
--- a/src/bypass.nim
+++ b/src/bypass.nim
@@ -1,0 +1,66 @@
+import std/[os, osproc, parseopt, strformat, strutils]
+
+type
+  bypassCmd = enum
+    bcNone
+    bcDocker = "docker"
+    bcExec = "exec"
+
+proc err(msg: string) =
+  ## Writes `msg` to stderr, then quits with an exit code of 1.
+  stderr.writeLine &"error: {msg}"
+  quit 1
+
+proc handleBypass*() =
+  ## Does nothing when the `CHALK_BYPASS` environment variable is unset.
+  ##
+  ## However, if that environment variable is set, and the command is:
+  ##
+  ## - `docker`: executes the rest of the command line with docker, without chalk.
+  ##
+  ## - `exec`: executes the rest of the command line, without chalk.
+  ##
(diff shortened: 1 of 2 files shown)


viega:
Per my last comment on the linked issue, I really don't think we should do this before the con4m refactor without a very good reason.

miki725:
+1. this approach I think breaks a bunch of semantics about chalk comment. if we need to bypass chalk, I think we need to do that in such a way which bypasses error-prone areas of chalk such as anything to do with user configs/reporting/etc without breaking user semantics

for example `chalk exec` needs to grab some potential exec args from either command line args or embedded config. saying that `CHALK_BYPASS` now breaks some chalk commands depending on context I dont think is acceptable.

a lot of con4m evaluation will be nop in v2 so over there we can correctly skip just the bits in chalk which can possible blow up such as chalk collection/reporting/etc.

miki725:
closing as we wont be taking this approach in the long term and this just pollutes PRs list

Issue #225, opened 2024-02-28
bug: chalk hanging on chalking jar files with cosign


indecisivedragon:
## Description

Initially reported by @thc202, debugging with @miki725 and @indecisivedragon. @miki725 was unable to reproduce locally but I was.

Running `chalk insert` on a `.jar` file with cosign enabled results in chalk hanging, suspected cause is file descriptor issue on read but not sure exactly.

## Impact

Chalking java projects is def broken, so holding up all the zap work. Also may be some bigger bugs in the underlying file descriptor code that we are passing around when we call cosign.

## Steps to Reproduce

1) build chalk latest
2) run `chalk setup` to generate keypairs (+ copy down the password)
3) `CHALK_PASSWORD=[password] ./chalk insert --trace --config-file=cosign.c4m test_zap/zap-2.15.0-SNAPSHOT.jar` (or whatever `.jar` file you happen to have handy. The files that I tested against were too large to be uploaded to github so ping me on slack if you need them because slack doesn't care.
4) observe chalk hangs

## Other Information

The arguments passed to cosign via chalk are:
```
trace: @["sign-blob", "--tlog-upload=false", "--yes", "--key", "chalk.key", "-"]
```
with blob:
```
trace: 72a4f5fb2eeaa82e24302ebe2d320a1f3016be091f7690f9e538a3a6755ca22f1514536cd37096634d9582923a094d543d9941da81b4858ca023c6a9b6fc3bfb
```
and default cosign location found in PATH. Calling `cosign` manually with this input does NOT cause cosign to fail or hang, so this is not a cosign error.

After some debugging, it looks like we are hanging somewhere in `sb_operate_switchboard` function in `nimutils/c/switchboard.c`, so this looks likely to be a nimutils error. The select statement seems to be returning a `-1` with `bad file descriptor` error.

PR #229, opened 2024-03-01
fix: chalking large zip files

miki725:

## Issue

none of the codecs were using chalk fd caching machinery and some of them were leaking FDs which was causing issues when trying to chalk a jar file with ~3K files

fixes https://github.com/crashappsec/chalk/issues/225
fixes https://github.com/crashappsec/chalk/issues/230

## Description

ensure consistent use of the FD cache mechanism across chalk codebase. previous each `ChalkObj` contained its own `stream: FileStream` attribute which was used for caching that paths FD. that had some issues such as it was impossible to use that mechanism outside of `ChalkObj` such as in codecs as they are responsible for create `ChalkObj` in the first place. to make it all consistent:

- [x] create generic fd_cache implementation independent of anything chalk-specific (eventually to be moved into nimutils)
- [x] refactor codecs to use generic `fd_cache` implementation
- [x] remove `stream` from `ChalkObj` and instead use the `fd_cache` from above
- [x] refactor existing chalk cache to use the `fd_cache`

## Testing

existing tests plus some tests on large zip files such as zap jar file which has ~3K files


src/commands/cmd_docker.nim
```diff
@@ -176,18 +177,18 @@ proc writeNewDockerFileIfNeeded(ctx: DockerInvocation) =
...
+  # add last blank line
   if ctx.inDockerFile.len() != 0 and ctx.inDockerFile[^1] != '\n':
     ctx.inDockerFile &= "\n"
-
   let newcontents = ctx.inDockerFile & ctx.addedInstructions.join("\n")
 
-  trace("New docker file: \n" & newcontents)
-  f.write(newcontents)
-  f.close()
+  let (f, path) = getNewTempFile()
```
viega, line 185: This should have an analogue `writeNewTempFile()` that does the stuff you do below.

src/docker_git.nim
```diff
@@ -54,8 +54,10 @@ proc createTempKnownHosts(data: string): string =
   if data == "":
     return ""
   let (f, path) = getNewTempFile()
-  f.write(data)
-  f.close()
+  try:
```
viega, line 57: See above, should have a 'writeNewTempFile()`; Plus, it occurs to me we should have a semgrep rule to warn on new adds of getNewTempFile()

src/fd_cache.nim
```diff
@@ -0,0 +1,310 @@
...
+
+##
+## FD cache
+## Global cache for opened files
+##
+## The overall concept is that there is a global cache
+## for all opened file streams.
+##
+## Some things you can do:
+## * yield   - create or get existing file stream from cache.
```
viega, line 16: 'yield' generally means "I'm done with this (for now)". Generally, I would take that to mean the system can decide whether it needs to be closed (whereas close generally asserts we need to close).   'Acquire' is the right word to use here.

src/fd_cache.nim
```diff
@@ -0,0 +1,310 @@
...
+##
+
+# TODO move this to nimutils
+
+import std/[enumerate, posix, streams, tables]
+import nimutils/file
+
+# ----------------------------------------------------------------------------
+
+proc getOpenLimit(): int =
```
viega, line 39: It may be worth limiting ourselves to some percentage of the rlimit (or subtract out a healthy number). Hard to know what 3rd party stuff we'll pull in that can't use the API.

src/plugins/codecMacOs.nim
```diff
@@ -55,135 +54,131 @@ template hasMachMagic(s: string): bool =
...
-      fullpath   = fullpath.replace("_CHALKSP_", " ")
-      trace("Will look for chalk mark in wrapper script: " & fullpath)
-      wrapStream = newFileStream(fullpath)
-
-      if wrapStream == nil:
-        warn("Previously chalked binary is missing its script. " &
-          "Replace the script or rename the executable")
-        scanFail()
-      # Drop down below for the chalk mark.
+  # chalked mac binary is a macho binary wrapped as shell script
```
viega, line 58: Mach-O is not macho 🤣

src/plugins/codecZip.nim
```diff
@@ -79,99 +77,93 @@ proc zipScan*(self: Plugin, loc: string): Option[ChalkObj] {.cdecl.} =
...
-        cache.onDisk.contents[k]  = newV
-      cache.embeddedChalk = collectionCtx.report
+      # Make sure the file seems to be a valid ZIP file.
+      var buf: array[4, char]
+      discard stream.peekData(addr(buf), 4)
+      if buf[0] != 'P' or buf[1] != 'K':
+        return none(ChalkObj)
 
-  return some(chalk)
+    let
```
viega, line 91: Someday we should really use an in-memory option here :/ Not now of course.

viega (changes requested):
I think this is very well done. The only thing is that 'yield' no longer has a correct meaning, so I think to avoid future confusion, please do change the naming as I mention in other comments!~

src/plugins/codecMacOs.nim
miki725: rofl 😂

src/fd_cache.nim
miki725: yep. we already do. we still honor the `cache_fd_limit` config:

https://github.com/crashappsec/chalk/blob/caf6e274b6ed384e476511fd16a898a7245bbbd1/src/util.nim#L301

and it checks that value is less than 1/2 of the rlimit:

https://github.com/crashappsec/chalk/blob/caf6e274b6ed384e476511fd16a898a7245bbbd1/src/fd_cache.nim#L223-L234

src/fd_cache.nim
miki725: was going between the 2 but yeah the logic for acquire makes more sense

src/commands/cmd_docker.nim
miki725: good idea. added in https://github.com/crashappsec/nimutils/commit/c5235862c40461068a7e929a9487dc9ab27d2495

src/docker_git.nim
miki725: cc @indecisivedragon was just mentioning semgrep to me as well. good idea to add that to chalk CI

src/fd_cache.nim
miki725: renamed in https://github.com/crashappsec/chalk/pull/229/commits/411b868deda805fca743cd1511845ad9adc44662

src/fd_cache.nim
miki725: also for context default fd limit in the config is 50:

https://github.com/crashappsec/chalk/blob/938e4746a852dbfbf75337f8b82451bc98c22af6/src/configs/chalk.c42spec#L2524-L2527

so we should be ok for most systems. dont imagine well see anything much lower with rlimit of 1024

viega (approved):

Issue #214, opened 2024-02-26
con4m v2 prep: stop using autogenerated con4m API


miki725:
currently con4m autogenerates a bunch of nim code in `src/c4autoconf.nim` which we use to reference con4m config values across the codebase. 

we wont be autogenerating same config utils in con4m v2 https://github.com/crashappsec/con4m/tree/jtv/v2. To help to fascilitate migration to v2, we can migrate all autogenerated calls to use con4m getters/settings now which will eventual migration to v2.

for example places like

https://github.com/crashappsec/chalk/blob/72a92cc3a4d3bb1beb03606e7a8d0863f88225e8/src/selfextract.nim#L301

will need to migrate to direct con4m getters. dont remember exact API as of now but it will need to use existing con4m getters:

https://github.com/crashappsec/con4m/blob/fad5f9719f6123d0c587c5424a8876eca410be64/files/con4m/st.nim#L207-L268

once this is refactored it should make migration to v2 a lot easier as hopefully well be able to batch replace most getters/setters to v2 API as well as change all the relevant con4m definitions in chalk

PR #235, opened 2024-03-07
refactor: begin to avoid autogenerated getters

ee7:
## Issue

https://github.com/crashappsec/chalk/issues/214

## Description

To ease the transition to con4m v2, we want to reduce our usage of autogenerated getters in `src/c4autoconf.nim`.

The all-encompassing `object` that `c4autoconf.nim` defines is `ChalkConfig`, which currently contains:

```nim
type ChalkConfig* = ref object
  `@@attrscope@@`*: AttrScope
  [...]
  attestationConfig*: AttestationConfig
  [...]
  extractConfig*: ExtractConfig
  dockerConfig*: DockerConfig
  execConfig*: ExecConfig
  loadConfig*: LoadConfig
  envConfig*: EnvConfig
  srcConfig*: SrcConfig
  cloudProviderConfig*: CloudProviderConfig
  [...]
```

As a starting point, try to refactor away all accesses of the fields that end in `Config` (which are shown above).

That is, with this PR, the following command no longer returns any matches:

```shell
git grep 'chalkConfig\.\w*Config\.'
```

I'll split this work across a few PRs. Another PR will be for all the fields that are of a simple built-in type.

In con4m v2, we'll be using [this proc](https://github.com/crashappsec/con4m/blob/94437931f2b1d5bb401c87320bc4525dd77e710d/src/attrstore.nim#L237-L249):

```nim
proc lookup*[T](ctx: RuntimeState, key: string): Option[T]
```

like

```nim
runtime.lookup("foo.bar").get()
```

so this PR makes it easier for us to refactor to that later.

This PR adds a `get` and `getOpt` in `config.nim` that calls the [`get` proc in con4m](https://github.com/crashappsec/con4m/blob/4bc04f1d4a725bdc61334e7423dc83b0d40c4d8f/files/con4m/st.nim#L252-L256) so we can write:

```nim
get[T](chalkConfig, "foo.bar")
```

or with the [`:T` syntax](https://nim-lang.github.io/Nim/manual.html#procedures-method-call-syntax) (edit: this was vetoed):

```nim
chalkConfig.get[:T]("foo.bar")
```

Refs: https://github.com/crashappsec/chalk/issues/214

## Testing

Hopefully the existing tests are sufficient.

### DIFF
--- a/src/attestation/backup.nim
+++ b/src/attestation/backup.nim
@@ -117,12 +117,11 @@ proc restore(self: Backup,
 proc initCallback(this: AttestationKeyProvider) =
   let
     self         = Backup(this)
-    backupConfig = chalkConfig.attestationConfig.attestationKeyBackupConfig
-    authName     = backupConfig.getAuth()
-    location     = backupConfig.getLocation()
+    authName     = get[string](chalkConfig, "attestation.attestation_key_backup.auth")
+    location     = get[string](chalkConfig, "attestation.attestation_key_backup.location")
     authOpt      = getAuthConfigByName(authName)
-    url          = backupConfig.getUri().removeSuffix("/")
-    timeout      = cast[int](backupConfig.getTimeout())
+    url          = get[string](chalkConfig, "attestation.attestation_key_backup.uri").removeSuffix("/")
+    timeout      = cast[int](get[Con4mDuration](chalkConfig, "attestation.attestation_key_backup.timeout"))
 
   if authOpt.isNone():
     raise newException(ValueError,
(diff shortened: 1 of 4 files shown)


src/config.nim
```diff
@@ -119,3 +120,9 @@ proc getPluginConfig*(name: string): Option[PluginSpec] =
 
 var autoHelp*:       string = ""
 proc getAutoHelp*(): string = autoHelp
+
+proc get*[T](chalkConfig: ChalkConfig, fqn: string): T =
+  get[T](chalkConfig.`@@attrscope@@`, fqn)
+
+proc getOpt*[T](chalkConfig: ChalkConfig, fqn: string): Option[T] =
+  getOpt[T](chalkConfig.`@@attrscope@@`, fqn)
```
miki725, line 128: 👍

miki725 (approved):

Issue #249, opened 2024-03-20
techStackGeneric: should be faster


ee7:
## Description

The [techStackGeneric plugin](https://github.com/crashappsec/chalk/blob/8f00241d0ae466af9601912e9f5e2bac6e30f22f/src/plugins/techStackGeneric.nim) is currently a bit too slow. It'd be good to speed it up. For example:

- [x] Profile it
- [x] Fix bugs in `getProcNames` (#250)
- [x] Stop reading files in `.git` (#253)
- [x] Don't hang on paths like `/dev/random` (https://github.com/crashappsec/chalk/pull/286/commits/38b75ebd048c66f28d615e92fb02d2b8886bd41c)
- [ ] Fix `scanDirectory` reading each file multiple times
- [ ] Stop scanning binaries
- [ ] Fix directory walking (symlink handling, prevent possible stack overflow, etc)
- [ ] Refactor away the unnecessary heap allocations
- [ ] Refactor away the globals
- [ ] Optimize the file reads. For example: consider using mmap
- [ ] Consider whether it's worth moving from `std/re` to e.g. [`nitely/nim-regex`](https://github.com/nitely/nim-regex). That's a pure Nim regex library which can be faster than PCRE for some types of regular expressions. It claims:

> The match time is linear in the length of the text and the regular expression. So, it can handle input from untrusted users. The syntax is similar to PCRE but lacks a few features that can not be implemented while keeping the space/time complexity guarantees, ex: backreferences.

For now, I don't think we'd need to parallelize it.

For a release build of chalk, profiling `chalk insert --use-tech-stack-detection foo` in a tiny repo with callgrind:

| Percentage of total execution time | Module | Procedure |
| -: | - | - |
| 90% | `techStackGeneric` | `scanFile` |
| which is due to: |  |  |
| 38% | `std/streams` | `readLine` |
| 22% | `fd_cache` | `acquireFileStream` |
| 18% | `std/re` | `find` |

## Dependencies

None.

## Subtickets

None.

PR #250, opened 2024-03-20
refactor(techStackGeneric): improve getProcNames

ee7:
### Description

The `getProcNames` procedure was doing a couple of strange things:

- It was trying to read files at `/proc/foo/status`, where `foo` is the name of a *file*, not directory, in `/proc`.
- It scanned each file at `/proc/[pid]/status` more than once (instead, once per digit of the pid).

This wasn't producing any incorrect results because:

- It caught and ignored the exceptions from opening nonexistent files.
- The name values are added to a `HashSet[string]`, which deduplicates them.

It was also doing some further unnecessary work: it kept scanning lines of `/proc/[pid]/status` even after it found the `Name` value (which is specified to be on the first line).

This PR fixes those issues.

Note that `/proc/[pid]/comm` also [exists in Linux since 2.6.33](https://www.man7.org/linux/man-pages/man5/proc.5.html) (2010-02-24), but that's less portable.

Refs: #249 (to a tiny degree)

### Testing

Add to `plugins/techStackDetection.nim`

```nim
when isMainModule:
  echo getProcNames()
```

and run `nim r plugins/techStackDetection.nim`

It'll be easier to add tests for this kind of thing when we're running unit tests in CI.

### DIFF
--- a/src/plugins/techStackGeneric.nim
+++ b/src/plugins/techStackGeneric.nim
@@ -176,22 +176,29 @@ proc hostHasTechStack(scope: hostScope, proc_names: HashSet[string]): bool =
 proc scanDirectory(directory: string, category: string, subcategory: string) =
   if inFileScope[category][subcategory]:
     return
-  for kind, path in walkDir(directory):
+  for filePath in walkDir(directory):
     if inFileScope[category][subcategory]:
       break
-    if kind == pcFile:
-      scanFile(path, category, subcategory)
-    elif kind == pcDir:
-      scanDirectory(path, category, subcategory)
+    if filePath.kind == pcFile:
+      scanFile(filePath.path, category, subcategory)
+      continue
+    if filePath.kind == pcDir:
+      scanDirectory(filePath.path, category, subcategory)
+      continue
 
 proc getLanguages(directory: string, langs: var HashSet[string]) =
-  for kind, path in walkDir(directory):
-    if kind == pcFile:
-      let ext = path.splitFile().ext
-      if ext != "" and ext in languages:
-        langs.incl(languages[ext])
-    elif kind == pcDir:
-      getLanguages(path, langs)
+  for filePath in walkDir(directory):
+    if filePath.kind == pcFile:
+      let splFile = splitFile(filePath.path)
+      if splFile.ext == "":
+        continue
+      if splFile.ext notin languages:
+        continue
+      langs.incl(languages[splFile.ext])
+      continue
+    if filePath.kind == pcDir:
+      getLanguages(filePath.path, langs)
+      continue
 
 proc detectLanguages(): HashSet[string] =
   result = initHashSet[string]()


nettrino (approved):

src/plugins/techStackGeneric.nim
```diff
@@ -123,20 +123,21 @@ proc scanFile(filePath: string, category: string, subcategory: string) =
...
   for kind, path in walkDir("/proc/"):
-    for ch in path.splitPath().tail:
-      try:
-        if ch notin "0123456789":
+    if kind == pcDir and path.lastPathPart().allIt(it in {'0'..'9'}):
+      let data =
+        try:
+          readFile(path / "status")
+        except:
           continue
```
miki725, line 135: We have a utility in nimutils tryToLoadFile. Let's switch to that so chalk is consistent with file reads. Once the fd_cache will move to nimutils we can adjust it to use the cache as well

src/plugins/techStackGeneric.nim
```diff
@@ -123,20 +123,21 @@ proc scanFile(filePath: string, category: string, subcategory: string) =
...
-        let p_path = path / "status"
-        var data = p_path.readFile()
-        for line in data.split("\n"):
-          if "Name:" in line:
-            var name = line.split("Name:")[1].strip()
-            result.incl(name)
-      except:
-         continue
+      for line in data.split("\n"):
```
miki725, line 136: Why not splitLines stdlib function?

miki725: You can also read the file one line at a time if you use the file stream from fd_cache so that we don't need to read the full file and then split by lines

src/plugins/techStackGeneric.nim
```diff
@@ -178,29 +179,22 @@ proc hostHasTechStack(scope: hostScope, proc_names: HashSet[string]): bool =
 proc scanDirectory(directory: string, category: string, subcategory: string) =
   if inFileScope[category][subcategory]:
     return
-  for filePath in walkDir(directory):
+  for kind, path in walkDir(directory):
     if inFileScope[category][subcategory]:
       break
```
miki725, line 184: Not following logic here. Why break out of the inner loop on a variable which is not controlled by the loop? Can we bypass loop if the condition doesn't match?

src/plugins/techStackGeneric.nim
```diff
@@ -178,29 +179,22 @@ proc hostHasTechStack(scope: hostScope, proc_names: HashSet[string]): bool =
...
-    if filePath.kind == pcFile:
-      scanFile(filePath.path, category, subcategory)
-      continue
-    if filePath.kind == pcDir:
-      scanDirectory(filePath.path, category, subcategory)
-      continue
+    if kind == pcFile:
```
miki725, line 185: Not sure what the nim conventions are but maybe use switch here as it switched on the kind and there are no other conditions in the branches?

miki725 (changes requested):

miki725 (approved):

PR #254, opened 2024-03-22
build: pinning cosign

miki725:

## Issue

tests are failing

https://github.com/sigstore/cosign/issues/3620

## Description

looks like hotfixes to cosign v1 are published as :latest tags and there is no v2 tag for grabbing latest v2 cosign versions so pinning to latest v2 for now

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 FROM ghcr.io/crashappsec/nim:ubuntu-2.0.0 as nim
-FROM gcr.io/projectsigstore/cosign as cosign
+FROM gcr.io/projectsigstore/cosign:v2.2.3 as cosign
 
 # -------------------------------------------------------------------
 


miki725:
dont think we should assume malice as we use cosign as one of the primary deps in chalk. docker tags are standard way of pinning images in docker ecosystem where you can easily understand what version it corresponds to whereas pinning by digest is less readable and is more complicated for things like multi-platform builds. for example if you pull image locally and grab its digest it will be wrong digest. so I dont think docker things should be pinned by digest unless you have a strong reason to do so which we dont have here

miki725:
as mentioned previously I dont put issue references in commit messages as links can change over time due to repo renames/etc. if viewing commits in github ui it shows which PR merged that commit so its easy enough to trace commit back to a PR which links to an issue

PR #257, opened 2024-03-27
fix: docker command name is honored while wrapping docker exe

miki725:

## Issue

unable to execute docker command which has "docker" as an arg somewhere in the params

https://github.com/zaproxy/zaproxy/actions/runs/8452961393/job/23155525385#step:9:482

## Description

previously "docker" was only wrapped when "docker" was not present anywhere in the flags. That is a bad assumption as for example a valid docker context can be a folder "docker". Right now if the exe name is "docker" its always appended to the CLI params hence allowing chalk to correctly validate CLI args.

## Testing

```
➜ make tests args="test_docker.py::test_docker_context --logs"
```

### DIFF
--- a/src/confload.nim
+++ b/src/confload.nim
@@ -177,14 +177,11 @@ proc loadAllConfigs*() =
   let
     toStream = newStringStream
     stack    = newConfigStack()
+    exeName  = getMyAppPath().splitPath().tail
 
-  case getMyAppPath().splitPath().tail
+  case exeName
   of "docker":
-    if "docker" notin params:
-      if len(params) != 0 and params[0] == "chalk":
-        params = params[1 .. ^1]
-      else:
-        params = @["docker"] & params
+    params = @[exeName] & params
   else: discard
 
   con4mRuntime = stack
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@
   installed. Previously, any chalk sub-scan such as
   during `chalk exec` had misleading error logs.
   [#248](https://github.com/crashappsec/chalk/pull/248)
+- Fixed parsing CLI params when wrapping `docker`
+  (rename `chalk` exe to `docker`) and a docker command
+  had a "docker" param.
+  [#257](https://github.com/crashappsec/chalk/pull/257)
 
 ## 0.3.4
 
--- a/src/util.nim
+++ b/src/util.nim
@@ -375,11 +375,14 @@ proc findExePath*(cmdName:    string,
       let
         subscan   = runChalkSubScan(location, "extract")
         allchalks = subscan.getAllChalks()
-      if len(allChalks) != 0 and allChalks[0].extract != nil and
-         "$CHALK_IMPLEMENTATION_NAME" in allChalks[0].extract:
-        continue
-      else:
+        isChalk   = (
+          len(allChalks) != 0 and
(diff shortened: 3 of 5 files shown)


src/confload.nim
```diff
@@ -177,14 +177,11 @@ proc loadAllConfigs*() =
...
     toStream = newStringStream
     stack    = newConfigStack()
+    exeName  = getMyAppPath().splitPath().tail
 
-  case getMyAppPath().splitPath().tail
+  case exeName
   of "docker":
-    if "docker" notin params:
-      if len(params) != 0 and params[0] == "chalk":
-        params = params[1 .. ^1]
```
miki725, line 185: @viega what was this guarding against? I cant think of a case where this can apply. if your executable name is `docker` when would `chalk` be the first parameter?

viega: Indirect recursion due to multiple chalks lying around.

viega (approved):

src/confload.nim
miki725: did a bunch of experiments locally and dont think recursion applies here, at least not anymore

there are a couple of cases where recursion can happen:

* `chalk docker ...` or `cp chalk docker && docker ...` (which is chalk internally). in this case the actual `docker` to be executed is found via `findExePath` which will found exe on `PATH` but will ignore all other chalks so recursion should not be possible there
  https://github.com/crashappsec/chalk/blob/b8b3481d53374d5bc5d251594b47c4708d54ebd7/src/util.nim#L355
  verified with a nested recursion with (excuse some of the fish shell conventions):
  ```
  ➜ PATH=$PWD/one:$PWD/two:$PWD/three:(echo $PATH | tr ' ' ':') ./docker version
  ```
  that correctly avoided all chalk dockers in `one`/`two`/`three` although noticed subscan was used on all dockers on PATH even after non-chalk docker was found. as subscan is kind of expensive addressed that in [`b8b3481` (#257)](https://github.com/crashappsec/chalk/pull/257/commits/b8b3481d53374d5bc5d251594b47c4708d54ebd7)
* mac codec does strange things due to a bash script wrapper for chalk. in that case that also works as expected as underlying binary name stays unchanged.
  https://github.com/crashappsec/chalk/blob/b8b3481d53374d5bc5d251594b47c4708d54ebd7/src/plugins/codecMacOs.nim#L18
  https://github.com/crashappsec/chalk/blob/b8b3481d53374d5bc5d251594b47c4708d54ebd7/src/plugins/codecMacOs.nim#L30
  so if you `cp chalk docker && ./docker` the underlying executable will still be called `docker` from chalk perspective and so same logic as above applies
* for all other cases recursion should be picked up specifically by recursion check in chalk:
  https://github.com/crashappsec/chalk/blob/b8b3481d53374d5bc5d251594b47c4708d54ebd7/src/norecurse.nim#L16-L38
* final case I can think of is `chalk exec -- chalk exec -- echo foo` but I dont think we ever addressed what we actually want to do there. do we run ignore inner chalk there or skip it and execute inner command. there are arguments in both directions. but either way dont think this case applies here as the exe name wont be `docker` there and the if statement here was only protection `./docker ...`

I currently cant compile on intel mac so cant verify 100% its working legit on mac but I did some POCs with similar bash scripts and can verify exe name wasnt changed by the mac codec

should we keep it simple here and if it causes any issues we can address later with a comment explaining the rational for the fix?

Issue #142, opened 2024-01-22
Consider pinning CI dependencies more strictly


ee7:
Currently, this is possible:

1. CI passes on a PR
2. We merge the PR
3. CI fails on `main`

I suggest that we should minimize the possibility of this because:

- CI should always be passing on `main`
- PRs are harder to review when there are unrelated CI failures
- It causes a context switch - CI should be fixed before other PRs can be merged

Recently, for example, CI passed on PR https://github.com/crashappsec/chalk/pull/135 but then failed on `main` with https://github.com/crashappsec/chalk/commit/f496958f9e3078f7f52b5e84bfdce5847dee6f26. @miki725 says it's because a new version of semgrep was released. This also caused CI to fail in PR https://github.com/crashappsec/chalk/pull/137.

In this case, can we pin the versions of dependencies like semgrep, and have automated PRs to update them?

Alternatively, the problem is alleviated somewhat if we use a merge queue.

miki725:
We actually already do that:

All test deps are pinned in https://github.com/crashappsec/chalk/blob/main/tests/pyproject.toml
All system test deps are pinned in https://github.com/crashappsec/chalk/blob/main/tests/Dockerfile

`semgrep` is test is special as `chalk` itself installs it as per `sast_config` during `insert/build` operations:

https://github.com/crashappsec/chalk/blob/991a87f989965994bede3539e793b859ff1ced47/src/configs/sastconfig.c4m#L45-L68

As such tests indirectly install it via chalk which I think is good as it tests whether chalk can correctly install/use the tool as per the config. The bigger issue was that the test was asserting full output payload of the semgrep tool which included its version. At the time when the test case was added in https://github.com/crashappsec/chalk/pull/129 `ContainsMixin` did not have the capability to correctly handle nested lists which semgrep output had which is why full payload was asserted (@indecisivedragon brought that up outside of GitHub). As part of fixing the `main`, now we now have the capability to do partial assertions including on lists and therefore any similar tests in the future can be written better from the start.

Not sure if you have anything else in scope for this issue?

miki725:
closing for now. feel free to reopen if the issue was not answered/addressed

PR #259, opened 2024-03-28
ci: use ubuntu-22.04, not ubuntu-latest

ee7:
The `ubuntu-latest` label [currently uses Ubuntu 22.04][1], but will likely use Ubuntu 24.04 later this year. Pin the Ubuntu version so that our upgrading is explicit.

Cosign being unpinned (https://github.com/crashappsec/chalk/pull/254) caused some unrelated CI failures recently, and I want to minimize that class of problems. I argue that the increased robustness from this PR is worth the cost of handling upgrades (where PRs can be created automatically).

The `ubuntu-latest` label started using Ubuntu 22.04 at the end of 2022, but the rollout [was advertised to take 8 weeks][2]. This PR prevents possible complications due to the runner image changing significantly from one workflow run to the next, even with the same repo state.

Note that the Ubuntu 22.04 image is still updated over time.

See some example differences between the installed and default versions of tools on different images:

- [`Ubuntu2004-Readme.md`](https://github.com/actions/runner-images/blob/07d23cf65fdc/images/ubuntu/Ubuntu2004-Readme.md)
- [`Ubuntu2204-Readme.md`](https://github.com/actions/runner-images/blob/07d23cf65fdc/images/ubuntu/Ubuntu2204-Readme.md)

Refs: https://github.com/crashappsec/chalk/issues/142

[1]: https://github.com/actions/runner-images/blob/07d23cf65fdc/README.md#available-images
[2]: https://github.blog/changelog/2022-11-09-github-actions-ubuntu-latest-workflows-will-use-ubuntu-22-04/

### DIFF
--- a/src/attestation/backup.nim
+++ b/src/attestation/backup.nim
@@ -117,12 +117,11 @@ proc restore(self: Backup,
 proc initCallback(this: AttestationKeyProvider) =
   let
     self         = Backup(this)
-    backupConfig = chalkConfig.attestationConfig.attestationKeyBackupConfig
-    authName     = backupConfig.getAuth()
-    location     = backupConfig.getLocation()
+    authName     = get[string](chalkConfig, "attestation.attestation_key_backup.auth")
+    location     = get[string](chalkConfig, "attestation.attestation_key_backup.location")
     authOpt      = getAuthConfigByName(authName)
-    url          = backupConfig.getUri().removeSuffix("/")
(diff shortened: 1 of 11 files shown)


miki725:
all the testing is done via docker so there is little risk to keep using latest ubuntu but I dont have a strong opinion on that. if we want to pin, Im ok with that

miki725 (approved):

PR #270, opened 2024-04-11
refactor: docker inspection refactor (ee7 review)

ee7:
Suggestions for #266.

I'll do this via PR in this case because:

- I review locally anyway
- I make the changes locally anyway
- I write the commit messages anyway
- It ensures that commits in the PR contain the review comments
- I don't have to copy changes and comments to the PR via the web UI, or use a tool that supports applying local changes to a PR via the GitHub API
- I can check that the changes compile, and make fewer broken suggestions

See commit messages.

---

To-do:

- [ ] Finish reviewing https://github.com/crashappsec/chalk/pull/266/commits/84a01736f5b9246b8fd23e0d6a8f3835019d7c52


src/chalk_common.nim
```diff
@@ -279,10 +279,9 @@ type
   DockerManifestType* = enum
     list, image, config, layer
 
-  DockerPlatform* = tuple[
-    os:           string,
-    architecture: string,
-  ]
+  DockerPlatform* = object
+    os*:           string
+    architecture*: string
```
miki725, line 284: this was explicitly done as a tuple so that all comparisons are done correctly by nim. does nim do the right thing for `==` on objects as well?

src/util.nim
```diff
@@ -510,11 +510,11 @@ proc removeSuffix*(s: string, suffix: string): string =
   result = s
   result.removeSuffix(suffix)
 
-proc removePrefix*(s: string, suffix: string): string =
```
miki725, line 513: oh haha. copy-paste from above. forgot to change all the vars

src/util.nim
```diff
@@ -528,12 +528,13 @@ proc `&=`*(a: var JsonNode, b: JsonNode) =
     a.add(i)
 
 proc getStrElems*(node: JsonNode): seq[string] =
+  result = @[]
```
miki725, line 531: doesnt nim automatically init the result?

miki725:
some feedback is addressed in https://github.com/crashappsec/chalk/pull/282

Issue #288, opened 2024-05-06
build: debug builds shouldn't be stripped


ee7:
### Scenario

Run `nimble build -d:debug`

### Expected behavior

It produces a binary with debug info.

### Observed behavior

It produces a binary without debug info.

### Problem

We make `nimble build` produce a release build by default, requiring the user to pass `-d:debug` for a debug build:

https://github.com/crashappsec/chalk/blob/ca217bd77211a0a6125840fd9167e3f76340738b/config.nims#L9-L13

and in the nimble file we intend to strip only the release binary:

https://github.com/crashappsec/chalk/blob/ca217bd77211a0a6125840fd9167e3f76340738b/chalk.nimble#L78-L81

However the above doesn't work, because nimble doesn't support checking custom defines in the nimble file.

PR #289, opened 2024-05-06
build(nimble): no longer strip debug builds

ee7:
The post-build hook attempted to strip only release builds, but it stripped all builds because [Nimble doesn't support checking custom defines in a nimble file][1]. Check against `release` instead, which is built-in.

Before this PR:

```console
$ nimble build -d:debug
[...]
$ file ./chalk
./chalk: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, stripped
```

With this PR:

```console
$ nimble build -d:debug
[...]
$ file ./chalk
chalk: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, with debug_info, not stripped
```

Closes: https://github.com/crashappsec/chalk/issues/288

[1]: https://www.github.com/nim-lang/nimble/issues/605

## Testing

- run `nimble build -d:debug`, and verify that the binary is not stripped
- run `nimble build`, and verify that the binary is stripped

### DIFF
--- a/chalk.nimble
+++ b/chalk.nimble
@@ -76,7 +76,7 @@ before install:
 
 # Add --trace if needed.
 after build:
-  when not defined(debug):
+  when defined(release):
     exec "set -x && strip " & bin[0]
   exec "set -x && ./" & bin[0] & " --debug --no-use-external-config --skip-command-report load default"
 


miki725 (approved):

miki725:
already solved via env var as nimble doesnt seem to honor var definitions

PR #293, opened 2024-05-24
fix: fetch GCP metadata from 169.254.169.254 directly

### LINKED ISSUE #292: Change metadata ip for GOOGLE to use IP
## Description

There might be services such as cloud run where it is not possible to fetch metadata from `metadata.google.internal` but it is doable via 169.254.169.254

## Impact

No cloud metadata is received for some cloud services

nettrino:

## Issue

Fixes #292

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,9 @@
 - Docker wrapping was resetting image `CMD` when base
   image had `ENTRYPOINT` defined.
   [#286](https://github.com/crashappsec/chalk/pull/286)
+- GCP instance metadata collection does not work by DNS
+  name reliably so switched to hard-coded IP.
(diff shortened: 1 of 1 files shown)


nettrino:
> We should probably define `169.254.169.254` as a `const` in this file and use that everywhere, but I don't mind leaving that till later.

Agreed - not sure though if there are lists of IPs that differ slightly per provider? this plugin could be split / refactored in the future so left it as such for now as I treated it as a nit as well

thc202:
Alibaba Cloud uses a different address: https://www.alibabacloud.com/help/en/ecs/user-guide/view-instance-metadata (not arguing against or for, just providing an example).

miki725 (approved):

Issue #298, opened 2024-05-28
vendor information not always present in providers, although cloud metadata endpoint is 


nettrino:
## Description

Cloud metadata might be available by querying the respective endpoint (e.g., in cloud functions) but the respective vendor information is not available, therefore no metadata gets populated by the cloud metadata plugin

## Impact

Metadata is missed 

## Steps to Reproduce

Deploy an image built by chalk in GCP cloud run and execute it

PR #311, opened 2024-05-31
feat: add _GCP_PROJECT_METADATA key

nettrino:

## Issue

Additional improvement over https://github.com/crashappsec/chalk/issues/298

## Description

Project information is not included in the instance metadata, therefore this exposes a separate key `_GCP_PROJECT_METADATA` with project info, if one is found

## Testing

Deploy a chalked image to GCP and check that project metadata is populated

### DIFF
--- a/src/plugins/cloudMetadata.nim
+++ b/src/plugins/cloudMetadata.nim
@@ -155,7 +155,7 @@ proc isAwsEc2Host(vendor: string): bool =
 
   return false
 
-proc isGoogleHost(vendor: string): bool =
+proc isGoogleHost(vendor: string, resolvContents: string): bool =
   # vendor is present
   if contains(strutils.toLowerAscii(vendor), "google"):
     return true
@@ -163,7 +163,6 @@ proc isGoogleHost(vendor: string): bool =
   # vendor information should be present in most services, but its not present
   # in cloud run. In cloud run we can detect the presence of a knative service
   # via ENV variables but we are being conservative in also checking resolv.conf
-  let resolvContents = tryToLoadFile("/etc/resolv.conf")
   var hasGoogleInternal = false
   for line in resolvContents.splitLines():
     # Checking that resolv.conf contains `google.internal` outside of a comment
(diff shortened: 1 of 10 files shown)


miki725 (changes requested):
Let's add some tests as well

nettrino:
> Let's add some tests as well

@miki725 we need to mock /etc/resolv.conf not sure what the best way is to go about this?

src/plugins/cloudMetadata.nim
```diff
@@ -199,15 +199,33 @@ proc cloudMetadataGetrunTimeHostInfo*(self: Plugin, objs: seq[ChalkObj]):
...
     isSubscribedKey("_OP_CLOUD_PROVIDER_TAGS") or
     isSubscribedKey("_OP_CLOUD_PROVIDER_ACCOUNT_INFO") or
     isSubscribedKey("_OP_CLOUD_PROVIDER_SERVICE_TYPE") or
     isSubscribedKey("_OP_CLOUD_PROVIDER_INSTANCE_TYPE")):
+    trace("Querying GCP metadata")
+    if isSubscribedKey("_GCP_PROJECT_METADATA"):
+      let projectOpt = hitProviderEndpoint("http://169.254.169.254/computeMetadata/v1/project/?recursive=true", newHttpHeaders([("Metadata-Flavor", "Google")]))
+      if projectOpt.isSome():
+        let valueProj = projectOpt.get()
+        try:
```
miki725, line 214: why 2 try excepts for a single chalk key?in the last comment I meant to move the `parseJson` as well as `nimJsonToBox` into a single `try..except`

tests/functional/test_plugins.py
```diff
@@ -498,6 +497,146 @@ def test_metadata_azure(
...
+    copy_files: list[Path],
+    chalk: Chalk,
+    tmp_file: Path,
+    server_imds: str,
+):
+    # make imds plugin think we are running in EC2
+    tmp_file.write_text("nothinghere")
+    bin_path = copy_files[0]
+    with open("/etc/resolv.conf", "a") as outf:
+        outf.write("search google.internal.\n")
```
miki725, line 512: this can interfere with other tests as the resolv file is global

for easy testing might be worth adding a config in chalk where it should search for resolv file. this way we can customize it below similar to how we override `VENDOR` which tells chalk the path of the vendor file

nettrino: yeah but a search entry should be benign for the most part? having a separate config option just for this would require additional changes? or could it be just a user defined bit? If we are not make any other chalk changes and support it via a config as it is let's do it that way

src/plugins/cloudMetadata.nim
nettrino: there were two different errors traced but can combine

tests/functional/test_plugins.py
miki725: tests run in parallel so adjusting resolv can impact other tests running at the same time. previously we had issues with leaking tests so if we are touching anything global, we should handle that appropriately

nettrino: can we do this just w/ a custom chalk config without any other chalk changes?

nettrino: It is either there or not in most cases though so making it a param just for the sake of a test feels like an overkill. Most other resolv.conf equivalents are just updating /etc/resolv.conf usually https://tailscale.com/blog/sisyphean-dns-client-linux. Even in parallel tests I don't think an extra `search ` entry would impact anything?

miki725: there are multiple concerns going on here:

* once we have multiple tests touching the same global dependency, they will all fight with each other. The worst part due to parallel tests it wont be deterministic which is why we had a bunch of issues like that before which were "fun" to figure out. Lets say azure implementation will also touch `resolv` file? Or well need multiple variants of GCP tests all running at the same time. They all will keep appending to the resolv file which wont be expected.
* currently we dont really support running tests directly from host but in eventually we should. Touching any host-level system things is not nice when running tests. Isolating it to a temporary file will ensure tests cannot possible impact system operation in anyway.

we already do that for the vendor file paths, even though their paths are also hardcoded per cloud provider but we use a config value to define the path allow:

* customize the path in case they run "interesting" setups like VMs
* easier testing

nettrino: exposed a `sys_resolv..` key under cloud_provider indentification, as it is used in this particular context

src/plugins/cloudMetadata.nim
```diff
@@ -155,15 +155,15 @@ proc isAwsEc2Host(vendor: string): bool =
 
   return false
 
-proc isGoogleHost(vendor: string): bool =
+proc isGoogleHost(vendor: string, resolv_file:string): bool =
```
miki725, line 158: nit:

```suggestion
proc isGoogleHost(vendor: string, resolv_file: string): bool =
```

tests/functional/test_plugins.py
```diff
@@ -498,6 +497,144 @@ def test_metadata_azure(
...
 
+@pytest.mark.parametrize("copy_files", [[LS_PATH]], indirect=True)
+@pytest.mark.parametrize("tmp_file", [{}], indirect=True)
+def test_metadata_gcp_no_vendor(
+    copy_files: list[Path],
+    chalk: Chalk,
+    tmp_file: Path,
+    server_imds: str,
+):
+    # make imds plugin think we are running in EC2
```
miki725, line 508: copy-paste? this is not anything EC2 related here that I can see

miki725 (commented):
👍 on the the conf for the resolv

some nit comments and looks like tests caught a regression for GCP when the vendor file is missing it doesnt set cloud provider key:

```
E               AssertionError: report['_OP_CLOUD_PROVIDER']: is missing
```

nettrino:
> 👍 on the the conf for the resolv
> 
> some nit comments and looks like tests caught a regression for GCP when the vendor file is missing it doesnt set cloud provider key:
> 
> ```
> E               AssertionError: report['_OP_CLOUD_PROVIDER']: is missing
> ```

probably test related as things were passing when I edited the file directly

miki725 (approved):

PR #315, opened 2024-06-04
fix: docker non-buildx file copy ensure name is unique

miki725:

## Issue

When docker context contained `docker` or `chalk` folders, entrypoint wrapping would fail

## Description

For non-buildx build, as the file needed to be copied into the context to be copied in Dockerfile, the name needs to be unique as if the same filename/folder already exist in the context folder, the copy/move will fail which in turn would bail out of ENTRYPOINT wrapping. Now chalk ensures that filename is unique for both copy/move modes.

In addition when using buildx, whenever chalk binary was located in a folder which contained .dockerignore which ignored the `chalk` binary (e.g. in chalk repo itself), the wrapping would fail as the binary could not be copied into the build. Now this edge case is detected and the file is copied into a temporary directory hence guaranteeing there is no .dockerignore there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,15 @@
   scans when the Docker context was outside that
   directory.
   ([#314](https://github.com/crashappsec/chalk/pull/314))
+- Docker wrapping `ENTRYPOINT` without `buildx` was failing
+  when docker context folder already had either `chalk`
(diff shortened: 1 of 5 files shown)


miki725:
> Are there any significant TOCTOU issues here?


no

PR #329, opened 2024-06-11
refactor: remove remaining chalkConfig field accesses 

ee7:

## Issue

#214

## Description

Continue from #235, #236, and #237, removing (nearly) every remaining access of a `ChalkConfig` field.

With this PR, the remaining field accesses of `chalkConfig` are the below:

```console
$ git log -1 --format='%h %s'
db26f26 refactor(all): avoid chalkConfig.(mark|report)Templates
$ git grep --break --heading --ignore-case 'chalkConfig\.'
src/run_management.nim
31:  get[T](chalkConfig.`@@attrscope@@`, fqn)
34:  getOpt[T](chalkConfig.`@@attrscope@@`, fqn)
```

Similar to previous PRs, I don't love this class of change, but we can add back some more compile-time safety around this later. The direction was to remove c4autoconf first.

## Testing

Hopefully the existing tests are sufficient.

### DIFF
--- a/src/config.nim
+++ b/src/config.nim
@@ -87,7 +87,7 @@ template forceReportKeys*(keynames: openarray[string]) =
     # Create the item section if required.
     if item notin getContents(keys):
       discard attrLookup(keys, [item], ix = 0, op = vlSecDef)
-    doAssert attrSet(keys, item & ".use", pack(true), Con4mType(kind: TypeBool)).code == errOk
+    con4mAttrSet(keys, item & ".use", pack(true), Con4mType(kind: TypeBool))
 
 template forceChalkKeys*(keynames: openarray[string]) =
   if isChalkingOp():
@@ -103,7 +103,7 @@ template forceChalkKeys*(keynames: openarray[string]) =
       # Create the item section if required.
       if item notin getContents(keys):
         discard attrLookup(keys, [item], ix = 0, op = vlSecDef)
-      doAssert attrSet(keys, item & ".use", pack(true), Con4mType(kind: TypeBool)).code == errOk
+      con4mAttrSet(keys, item & ".use", pack(true), Con4mType(kind: TypeBool))
 
 proc runCallback*(cb: CallbackObj, args: seq[Box]): Option[Box] =
   return con4mRuntime.configState.sCall(cb, args)
--- a/src/confload.nim
+++ b/src/confload.nim
@@ -98,7 +98,7 @@ proc loadLocalStructs*(state: ConfigState) =
       # no log as this function is called multiple times
       # and any logs are very verbose
       continue
-  doAssert state.attrSet("config_path", pack(configPath)).code == errOk
+  state.con4mAttrSet("config_path", pack(configPath))
   var c4errLevel =  if get[bool](chalkConfig, "con4m_pinpoint"): c4vShowLoc else: c4vBasic
 
   if get[bool](chalkConfig, "chalk_debug"):
--- a/src/run_management.nim
+++ b/src/run_management.nim
@@ -33,6 +33,21 @@ proc get*[T](chalkConfig: ChalkConfig, fqn: string): T =
 proc getOpt*[T](chalkConfig: ChalkConfig, fqn: string): Option[T] =
   getOpt[T](chalkConfig.`@@attrscope@@`, fqn)
 
+proc con4mAttrSet*(ctx: ConfigState, fqn: string, value: Box) =
+  ## Sets the value of the `fqn` attribute in `ctx.attrs` to `value`, raising
+  ## `AssertionDefect` if unsuccessful.
+  ##
+  ## This proc must only be used if the attribute is already set. If the
+  ## attribute isn't already set, use the other `con4mAttrSet` overload instead.
+  doAssert attrSet(ctx, fqn, value).code == errOk
+
+proc con4mAttrSet*(attrs: AttrScope, fqn: string, value: Box, attrType: Con4mType) =
+  ## Sets the value of the `fqn` attribute in `attrs` to `value`, raising
+  ## `AssertionDefect` if unsuccessful.
+  ##
+  ## This proc may be used if the attribute is not already set.
+  doAssert attrSet(attrs, fqn, value, attrType).code == errOk
+
 # This is for when we're doing a `conf load`.  We force silence, turning off
 # all logging of merit.
 proc startTestRun*() =
--- a/src/subscan.nim
+++ b/src/subscan.nim
@@ -40,7 +40,7 @@ proc runChalkSubScan*(location: seq[string],
   try:
     if suspendHost:
       suspendHostCollection()
-    doAssert runtime.attrSet("recursive", pack(true)).code == errOk
+    runtime.con4mAttrSet("recursive", pack(true))
     result                = pushCollectionCtx()
     case cmd
     # if someone is doing 'docker' recursively, we look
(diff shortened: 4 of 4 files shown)


src/plugin_api.nim
```diff
@@ -577,7 +577,7 @@ proc checkPlugin(plugin: Plugin, codec: bool): bool {.inline.} =
 proc getAllPlugins*(): seq[Plugin] =
   var preResult: seq[(int, Plugin)] = @[]
   for name, plugin in installedPlugins:
-    preResult.add((plugin.configInfo.getPriority(), plugin))
+    preResult.add (get[int](plugin.configInfo, "priority"), plugin)
```
miki725, line 580: nit
```suggestion
    preResult.add(get[int](plugin.configInfo, "priority"), plugin)
```

src/chalkjson.nim
```diff
@@ -442,12 +442,12 @@ proc extractOneChalkJson*(chalkData: string, path: string): ChalkDict =
...
 # Output ordering for keys.
 proc orderKeys*(dict: ChalkDict,
-                tplate: MarkTemplate | ReportTemplate): seq[string] =
+                tplate: AttrScope): seq[string] =
   var tmp: seq[(int, string)] = @[]
   for k, _ in dict:
-    var order = chalkConfig.keySpecs[k].normalizedOrder
-    if tplate != nil and k in tplate.keys:
-      let orderOpt = tplate.keys[k].order
+    var order = get[int](getChalkScope(), "keyspec." & k & ".normalized_order")
```
miki725, line 448: is this guaranteed to be present?

miki725: looks like it is https://github.com/crashappsec/chalk/blob/6b6c9c260a9c18cc46f25407b773da83e9dcd80f/src/configs/chalk.c42spec#L827-L829

src/collect.nim
```diff
@@ -41,23 +41,28 @@ proc canWrite(plugin: Plugin, key: string, decls: seq[string]): bool =
...
 
-proc registerKeys(templ: MarkTemplate | ReportTemplate) =
-  for name, content in templ.keys:
-    if content.use: subscribedKeys[name] = true
+proc registerKeys(templ: AttrScope) =
+  let keyOpt = getObjectOpt(templ, "key")
+  if keyOpt.isSome():
+    let key = keyOpt.get()
+    for name, content in key.contents:
+      if content.isA(AttrScope):
```
miki725, line 63: what happens if its not an attrscope. perhaps do an assertion that it is?

src/collect.nim
```diff
@@ -41,23 +41,28 @@ proc canWrite(plugin: Plugin, key: string, decls: seq[string]): bool =
...
+        let useOpt = getOpt[bool](content.get(AttrScope), "use")
+        if useOpt.isSome() and useOpt.get(): subscribedKeys[name] = true
```
miki725, line 65:
```suggestion
        if useOpt.isSome() and useOpt.get():
          subscribedKeys[name] = true
```

previously it was an easy to read condition as there was only one `if <condition>:` whereas now its multiple conditions which makes it harder to read whats actually getting done

p.s. in general Im not a fan of single-line conditions and prefer to indent the code block to make it visually different

src/config.nim
```diff
@@ -48,48 +48,62 @@ proc selfChalkSetSubKey*(key: string, subKey: string, subValue: Box) =
...
+
   for item in keynames:
-    if item in templateRef.keys:
-      templateRef.keys[item].use = true
-    else:
-      templateRef.keys[item] = KeyConfig(use: true)
+    # Create the item section if required.
+    if item notin getContents(keys):
+      discard attrLookup(keys, [item], 0, vlSecDef)
+    doAssert attrSet(keys, item & ".use", pack(true), Con4mType(kind: TypeBool)).code == errOk
```
miki725, line 90: these sets are all kind of really ugly to follow. can we use util template or something to do the same but to make api easier to use?

```suggestion
    con4mAttrSet(keys, item & ".use", pack(true))
```

internally it can get its type from `box.kind` so no need to pass `Con4mType` directly

src/config.nim
```diff
@@ -48,48 +48,62 @@ proc selfChalkSetSubKey*(key: string, subKey: string, subValue: Box) =
...
 
-  chalkConfig.reportTemplates[tmplName]
+  getObject(getChalkScope(), "report_template." & tmplName)
 
 template forceReportKeys*(keynames: openarray[string]) =
   let templateRef = getReportTemplate()
 
+  # Create the "key" section if required.
+  if getObjectOpt(templateRef, "key").isNone() and keynames.len > 0:
+    discard attrLookup(templateRef, ["key"], 0, vlSecDef)
```
miki725, line 82: can we use keyword args for `0` and `vlSecDef`? I have no idea what they mean in this context. same in other `attrLookup` usage which create a section

src/config.nim
```diff
@@ -112,12 +126,15 @@ proc setCommandName*(s: string) =
 proc getChalkRuntime*(): ConfigState      = con4mRuntime.configState
 proc getValidationRuntime*(): ConfigState = con4mRuntime.validationState
 
-proc getKeySpec*(name: string): Option[KeySpec] =
-  if name in chalkConfig.keyspecs: return some(chalkConfig.keyspecs[name])
+proc getKeySpec*(name: string): Option[AttrScope] =
+  for k, v in getChalkSubsections("keyspec"):
```
miki725, line 130: chalk has a lot of keys. no way to get these as a dict for O(1) lookup if the key is present?

src/config.nim
```diff
@@ -112,12 +126,15 @@ proc setCommandName*(s: string) =
...
+    if name == k:
+      return some(v)
 
-proc getPluginConfig*(name: string): Option[PluginSpec] =
-  if name in chalkConfig.plugins:
-    return some(chalkConfig.plugins[name])
+proc getPluginConfig*(name: string): Option[AttrScope] =
+  for k, v in getChalkSubsections("plugin"):
```
miki725, line 135: same as above although less critical as there are much fewer plugins

src/config.nim
miki725: in docker util it uses:

https://github.com/crashappsec/chalk/pull/329/files#diff-53ca1218705d3afa936ec476df42cf4fcd3f6d02a0a99aad528a1778d9ec808dR110

```
  if key notin getContents(getChalkScope().getObject("keyspec")):
```

src/reporting.nim
```diff
@@ -105,21 +105,24 @@ template doEmbeddedReport(): Box =
...
+    if getCommandName() notin useWhen and "*" notin useWhen:
       continue
     if topic == "audit" and not get[bool](chalkConfig, "publish_audit"):
       continue
     if len(sinkConfs) == 0 and topic notin ["audit", "chalk_usage_stats"]:
       warn("Report '" & topic & "' has no configured sinks.  Skipping.")
 
-    let templateToUse = chalkConfig.reportTemplates[spec.reportTemplate]
+    let reportTemplate = get[string](spec, "report_template")
+    let templateToUse = getObject(getChalkScope(), "report_template." & reportTemplate)
```
miki725, line 125: nit

```suggestion
    let
      reportTemplate = get[string](spec, "report_template")
      templateToUse = getObject(getChalkScope(), "report_template." & reportTemplate)
```

chalk predominatly uses this pattern for defining multiple vars at once

src/plugin_api.nim
miki725: this can confuse others. we use `()` predomindantly in chalk. please keep parens in

src/config.nim
miki725: internally the util will still pass the `Con4mType`, but the caller of the utility will just pass a `Box` and it can normalize internally as necessary

src/collect.nim
miki725: is it guaranteed to be `AttrScope`? just trying to avoid regressions

miki725 (approved):

PR #347, opened 2024-06-17
fix(git): commit id is correctly determined for annotated symbolic ref tags

miki725:

## Issue

wrong commit id is reported

## Description

`.git/HEAD` can be a symbolic ref to a branch/tag or directly pointer to a commit. When doing `git checkout <branch>` it normally sets it to symbolic ref however when doing `git checkout <tag>` it usually sets it directly to the tag's commit. However it is possible to update it via `git symbolic-ref HEAD <tag>`. In that case `HEAD` will point to annotated tag and as such we chalk simply opens the ref of that symbolic ref it is not guaranteed to be for a commit id. Now chalk loads the tag object and if it is an annotated tag, resolves commit id to be from the tags object pointer.

## Testing

```
➜ make tests args="test_git.py::test_repo[False-copy_files0-True-None-True-True] --logs -x"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## On the `main` branch
 
+### Fixes
+
+- Chalk did not extract correct commit ID for git repos
+  with `HEAD` being symbolic reference to an annotated tag.
+  This usually happens via `git symbolic-ref HEAD`.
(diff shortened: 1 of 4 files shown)


miki725:
all other keys are derived from the commit so if chalk cant figure out correct commit id, rest are descriptions of the commit so I think thats the most critical part to mention in the changelog but we can improve before release if we like

PR #357, opened 2024-06-24
fix(docker): fixing --metadata-file support

miki725:

## Issue

chalk build fails due to invalid `--metadata-file` flag.

## Description

In older docker versions, as `docker build` is not an alias to `docker buildx build`, `--metadata-file` flag is not supported. It only works when explicitly using `docker buildx build` in which case 0.6.0 is required which added that flag. Otherwise docker version needs to be >=22 which aliased `build` to `buildx build` as well as have supported buildx version.

No easy way of testing it in CI but tested locally various combination of versions.

## Testing

Was running something like:

```
➜ echo '
  FROM docker:22.06-rc
  COPY --from=docker/buildx-bin:0.6.0 /buildx /usr/lib/docker/cli-plugins/docker-buildx
  COPY --from=docker/buildx-bin:0.6.0 /buildx /usr/local/libexec/docker/cli-plugins/docker-buildx
  ' | docker build -f - . -t dockerbuildx
  and docker run -it --rm dockerbuildx docker --version
  and docker run -it --rm dockerbuildx buildx version
  and docker run -it --rm dockerbuildx build --help
  and docker run -it --rm dockerbuildx buildx build --help
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ## On the `main` branch
 
+### Fixes
+
+- Docker build `--metadata-file` flag is only added when
+  using `buildx >= 0.6.0`. In addition the flag is only added
+  when using `docker >= 22` as docker aliased `docker build`
+  to `docker buildx build` which allows to use buildx flags
+  in normal build command.
+  ([#357](https://github.com/crashappsec/chalk/pull/357))
+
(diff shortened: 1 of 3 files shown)


MyNameIsMeerkat (approved):
LGTM 👍

PR #362, opened 2024-07-08
fix(cloudMetadata): set _OP_CLOUD_PROVIDER earlier

ee7:
Previously, a chalk report would omit the `_OP_CLOUD_PROVIDER` key when:

- No other metadata key (for example, `_GCP_INSTANCE_METADATA` or `_OP_CLOUD_PROVIDER_IP`) was subscribed.
- The instance metadata service either couldn't be reached, or returned invalid data.

Set that key sooner when it's subscribed and the cloud provider is known.

For now, duplicate the `isGoogleHost` and `isAzureHost` calls to keep this PR small. Those duplicate checks are cheap, but we can remove them later - this file would benefit from a larger refactoring.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,14 @@
 
 ### Fixes
 
-- The `_OP_CLOUD_PROVIDER` key could previously be missing
-  when the instance metadata service couldn't be reached or
-  returned invalid data.
+- A chalk report would previously omit the `_OP_CLOUD_PROVIDER`
+  key when:
+
+  - No other instance metadata key (e.g. `_GCP_INSTANCE_METADATA`
+    or `_OP_CLOUD_PROVIDER_IP`) was subscribed.
+  - The instance metadata service couldn't be reached, or
+    returned invalid data.
+
   ([#362](https://github.com/crashappsec/chalk/pull/362))
 
 ## 0.4.7


src/plugins/cloudMetadata.nim
```diff
@@ -200,13 +200,18 @@ proc cloudMetadataGetrunTimeHostInfo*(self: Plugin, objs: seq[ChalkObj]):
...
     isSubscribedKey("_OP_CLOUD_PROVIDER_IP") or
     isSubscribedKey("_OP_CLOUD_PROVIDER_REGION") or
     isSubscribedKey("_OP_CLOUD_PROVIDER_TAGS") or
     isSubscribedKey("_OP_CLOUD_PROVIDER_ACCOUNT_INFO") or
     isSubscribedKey("_OP_CLOUD_PROVIDER_SERVICE_TYPE") or
     isSubscribedKey("_OP_CLOUD_PROVIDER_INSTANCE_TYPE")
   ):
+    isSubscribedKey("_OP_CLOUD_PROVIDER"):
+      # FIXME use enum
+      result["_OP_CLOUD_PROVIDER"] = pack("gcp")
```
miki725, line 213:
```suggestion
    # FIXME use enum
    result.setIfNeeded("_OP_CLOUD_PROVIDER", "gcp")
```

similar below

thats a more standard pattern across other plugins. this plugin specifically doesnt follow most of the conventions elsewhere

src/plugins/cloudMetadata.nim
```diff
@@ -200,13 +200,18 @@ proc cloudMetadataGetrunTimeHostInfo*(self: Plugin, objs: seq[ChalkObj]):
   if isGoogleHost(vendor, resolv) and (
     isSubscribedKey("_GCP_INSTANCE_METADATA") or
     isSubscribedKey("_GCP_PROJECT_METADATA") or
+    isSubscribedKey("_OP_CLOUD_PROVIDER") or
     isSubscribedKey("_OP_CLOUD_PROVIDER_IP") or
     isSubscribedKey("_OP_CLOUD_PROVIDER_REGION") or
     isSubscribedKey("_OP_CLOUD_PROVIDER_TAGS") or
     isSubscribedKey("_OP_CLOUD_PROVIDER_ACCOUNT_INFO") or
     isSubscribedKey("_OP_CLOUD_PROVIDER_SERVICE_TYPE") or
     isSubscribedKey("_OP_CLOUD_PROVIDER_INSTANCE_TYPE")
```
miki725, line 209: lets say only `_OP_CLOUD_PROVIDER` is subscribed but with this condition it will hitting metadata endpoint even if no other key is subscribed

I think it should be a separate conditional block outside this block which sets just the `_CLOUD_PROVIDER`. it will then keep the correct subscription semantics

same below

miki725 (changes requested):

miki725 (approved):

miki725 (approved):

nettrino (approved):

PR #365, opened 2024-07-10
feat: setting chalk_log_level to trace in connect.c4m

miki725:
- [ ] ~Updated `CHANGELOG.md` if necessary~ (not updating as this is only change part of `connect.c4m` and not the default chalk config

## Issue

If the chalk has an error its hard to debug now as you have to hunt down execution logs, if chalk is running with trace log level. Otherwise you are out of 

## Description

This will help with debugging as chalk report will include all trace logs as part of the report in `_OP_ERRORS` field. In the future might be worth renaming the field as `ERRORS` sounds dangerous even when it contains trace-level logs...

We should do a pass on anything potentially sensitive we might be logging to ensure the field does not leak anything security-sensitive.

## Testing

```
➜ cat /etc/chalk.c4m | grep log_level
chalk_log_level = "trace"
```

then run a few commands such as `chalk insert ls` to see log output being part of the report

### DIFF
--- a/configs/connect.c4m
+++ b/configs/connect.c4m
@@ -47,6 +47,7 @@ custom_report crashoverride {
   ~use_when:        ["insert", "build", "push", "exec"]
 }
 
+chalk_log_level                      = "trace"
 attestation.key_provider             = "get"
 docker.wrap_entrypoint               = true
 run_sbom_tools                       = true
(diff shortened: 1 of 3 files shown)


miki725:
https://app.shortcut.com/crashoverride-1/story/955/chalk-add-option-to-send-logs-in-the-report

PR #370, opened 2024-07-11
fix: cloud metadata timeout handing and guaranteeing _OP_CLOUD_* keys

miki725:

## Issue

some keys were not guaranteed to be present and timeout errors were not handled in the plugin

## Description

When making HTTP calls the plugin did not handle timeout exceptions and
as such on any exception complete plugin will bail and none of its keys
were populated such as `_OP_CLOUD_METADATA`. We were observing a bunch
of logs:

```
warn:  When collecting run-time host info, plugin implementation cloud_metadata threw an exception it didn't handle: Call to 'connect' timed out.
```

Looks like 250ms wasnt enough in AWS for metadata plugin to reliably hit
metadata endpoint so bumping it to 500ms.

---

metadata endpoint can fail due to timeout however in most cases we can
still determine service type by the presence of env vars plus the
hardware vendor file already told us the cloud provider so we can set
the key regardless of the output of metadata endpoint response.

---

in addition ensuring all exception handling provides useful trace log
message as otherwise debugging any of the errors is not-trivial

## Testing

hard to test with test suite some of these. mostly manual testing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,14 +5,15 @@
 ### Fixes
 
 - A chalk report would previously omit the `_OP_CLOUD_PROVIDER`
-  key when:
+  and `_OP_CLOUD_PROVIDER_SERVICE_TYPE` keys when:
 
   - No other instance metadata key (e.g. `_GCP_INSTANCE_METADATA`
     or `_OP_CLOUD_PROVIDER_IP`) was subscribed.
   - The instance metadata service couldn't be reached, or
     returned invalid data.
 
-  ([#362](https://github.com/crashappsec/chalk/pull/362))
+  ([#362](https://github.com/crashappsec/chalk/pull/362)
+  [#370](https://github.com/crashappsec/chalk/pull/370))
 
 - `_OP_ERRORS` was missing any logs/errors from plugins.
   The key was collected by the `system` plugin which
--- a/src/plugins/cloudMetadata.nim
+++ b/src/plugins/cloudMetadata.nim
@@ -67,7 +67,6 @@ proc getAzureMetadata(): ChalkDict =
       isSubscribedKey("_OP_CLOUD_PROVIDER_REGION") or
       isSubscribedKey("_OP_CLOUD_PROVIDER_TAGS") or
       isSubscribedKey("_OP_CLOUD_PROVIDER_ACCOUNT_INFO") or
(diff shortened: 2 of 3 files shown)


miki725:
https://app.shortcut.com/crashoverride-1/story/925

miki725:
the use of the template was intentional here. its purpose is to handle exceptions differently:

for example:

```
proc foo(x: string) =
  try:
    echo(x)
  except:
    echo("inside proc exception")
    echo(getCurrentExceptionMsg())

template bar(x: untyped) =
  try:
    echo(x)
  except:
    echo("inside template exception")
    echo(getCurrentExceptionMsg())

let a: seq[string] = @[]

try:
  foo(a[5])
except:
  echo("outside proc exception")
  echo(getCurrentExceptionMsg())

bar():
  a[5]
```

note how they are differently handled:

```
➜ nim c -r test.nim
outside proc exception
index out of bounds, the container is empty
inside template exception
index out of bounds, the container is empty
```

`proc` param is executed **before** proc call so exception blows up before `proc` is ever called so you need to have the exception handling outside of the `proc`.

on the other hand `template` is passed a code block which is then executed **inside** the template which allows it to wrap that code block around `try..except`. thats the whole point of using template blocks - to allow to pass code blocks vs values which is what `proc` is limited to

that is exactly what `trySetIfNeeded` is meant to do. it allows to try to set a key while computing its value can blow which will log appropriate trace log message if it does blow up.

consider the suggestion above of:

```
      result.trySetIfNeeded("_OP_CLOUD_PROVIDER_TAGS",          jsonValue["compute"]["tagsList"].nimJsonToBox())
```

That assumes both `compute` key is present and its json value is an object. then it assumes `tagsList` key is present in that. finally it assumes that key returns non-nil value as `nimJsonToBox` doesnt handle nil values. Previously the whole statement was wrapped with an exception block:

```
try:
  result.setIfNeeded("...", ...)
except:
  ...
```

which works fine but is a little verbose once you are setting a lot of keys as you need to ensure each one of those keys has correct exception message.

with the template it allows to make that a lot simpler. just pass a code block to it and it will get executed internally within an exception block and will log the exception for you:

```
result.trySetIfNeeded("..."):
   ....
```

alternatively to a template block it could be implemented by passing an anonymous function but I think thats less explicit:

```
result.trySetIfNeeded("...", () => ...)
```

as its much easier to miss `() =>` being part of the param. I think separate code block makes it explicit that code block is evaluated later and not inline.

Issue #251, opened 2024-03-21
Chalk reporting invalid chalk mark in autocomplete script


drraid:
## Description
using chalk on linux produces an error about invalid chalk mark in the autocomplete script (the file which chalk itself creates/manages)

```
another1@covm2:~$ ./chalk
info:  Installed bash auto-completion file to: /home/another1/.local/share/bash_completion/completions/chalk.bash
info:  Added sourcing of autocomplete to ~/.bash_completion file
info:  Script should be sourced automatically on your next login.
another1@covm2:~$ ./chalk
info:  Invalid chalk mark in autocompletion script. Updating.
info:  Installed bash auto-completion file to: /home/another1/.local/share/bash_completion/completions/chalk.bash
```

## Impact

This is mostly an annoyance but could represent a bigger issue. 

## Steps to Reproduce

See above use of chalk on x64 linux

numerical:
Looks like `ARTIFACT_VERSION` is missing from the bash autocomplete chalk. `autoCompleteFileCheck` thus doesn't know what the semver is and just replaces the file on every invocation. 

autocomplete magic:
https://github.com/crashappsec/chalk/blob/7863da4b63ee3cc80fd46e28273808d26f739203/src/autocomplete/default.bash#L171 

relevant line in util:
https://github.com/crashappsec/chalk/blob/7863da4b63ee3cc80fd46e28273808d26f739203/src/util.nim#L242

PR #384, opened 2024-07-29
build: Update ARTIFACT_VERSION in autocomplete files

numerical:
Updates the autocomplete files to reflect the most recent version

A future improvement could be to pull the version information in util.nim from the same place that the Makefile pulls it

- [ ] Followed the steps in the contributor's guide: https://crashoverride.com/docs/other/contributing#filing-the-pull-request
- [ ] PR title uses [semantic commit messages](https://nitayneeman.com/posts/understanding-semantic-commit-messages-using-git-and-angular/#fix)
- [ ] Filled out the template to a useful degree
- [ ] Updated `CHANGELOG.md` if necessary

## Issue

Chalk never updates the autocomplete files to insert the ARTIFACT_VERSION so continually replaces the file as it's used.
#251 

## Description

Updates util.nim and the autocomplete files with the correct version.
Also updates the Makefile with a new target to run the nimble task that will update the autocomplete files

## Testing

1) compile chalk
2) run any chalk command to install the new autocomplete file
3) run any chalk command to see that the autocomplete file is no longer copied

```
~/chalk ❯❯❯ ./chalk env
info:  Updating autocomplete script to version: 0.4.9
info:  Installed bash auto-completion file to: /home/numeral/.local/share/bash_completion/completions/chalk.bash
info:  /home/numeral/.local/chalk/chalk.log: Open (sink conf='default_out')
info:  Full chalk report appended to: ~/.local/chalk/chalk.log
[...]

~/chalk ❯❯❯ ./chalk env
info:  /home/numeral/.local/chalk/chalk.log: Open (sink conf='default_out')
info:  Full chalk report appended to: ~/.local/chalk/chalk.log
```


numerical:
One issue in this PR is that the current version is still hardcoded, I do have updates to pull in the current version with `getChalkScope` and use `parseVersion` to do the comparison if that would be better

miki725:
thanks for the PR!

> One issue in this PR is that the current version is still hardcoded, I do have updates to pull in the current version with getChalkScope and use parseVersion to do the comparison if that would be better

there is actually a module for that already:

https://github.com/crashappsec/chalk/blob/e9a6a11865d42db1348c46b5beb7b49e2b12c144/src/config_version.nim

although it returns a string, not a tuple so there might need to be some adjusting somewhere to make it all work

---

as a side note we should def add a step in our release process to chalk those autocomplete files so that released version is always chalked with the corresponding release number chalkmark

numerical:
```/home/numeral/chalk/src/config_version.nim(11, 19) Error: 'staticRead' can only be used in compile-time context
```

the autocheck happens during runtime so I guess `get[string](getChalkScope(), 'keyspec.CHALK_VERSION.value')` has to be used here, but I think it prob makes sense to add a `withoutSuffix` to semver to strip the `-dev` just in case

numerical:
Hey @miki725 sorry it took so long to update this PR, hacker summer camp had me running around way too much. I updated the PR to check the current chalk version in `util.nim` so now it compares the autocomplete file version vs the current chalk version rather than a hardcoded value. Should be ready to merge (assuming good review etc).

PR #409, opened 2024-08-16
feat: adding co/ configs for custom chalk profiles

miki725:
these will be pulled by dynamic generated component


configs/co/profile.c4m
miki725, line 1: currently this is not used due to param variables not being able to use in keyspec values but this should be fixed in libcon4m so we can update this component as necessary whenever that ships

viega (approved):

PR #410, opened 2024-08-19
refactor: vendor con4m dev branch

ee7:

## Description

Previously, Chalk depended on the con4m repo, which itself depended on the nimutils repo. Vendor the con4m dev branch source files, and make Chalk depend on nimutils directly for now.

I think that's easier at this point for the libcon4m integration, because:

- We want to migrate Chalk from con4m v1 to libcon4m, and we decided that we'll do without making the con4m repo use libcon4m. That is, we'll make Chalk depend on libcon4m directly (and I believe archive the con4m repo).
- There is functionality in the con4m repo that won't be implemented by libcon4m, so Chalk needs to adopt and adapt that functionality.

Clearly, there's a bunch of things that we could adapt/discard from this vendored code. But I think it's easiest for this PR to just vendor it with the exact minimum of changes.

And I believe the consensus was that we should also just move nimutils source files into Chalk too, but let's do that in a separate PR too.

## Testing

This is expected to be refactoring-only, so hopefully it's sufficient for the existing tests to pass.

The reviewer might like to make this change themselves, and verify that this PR vendors the correct state with no changes.

### DIFF
--- a/chalk.nimble
+++ b/chalk.nimble
@@ -11,7 +11,7 @@ bin           = @["chalk"]
 
 # Dependencies
 requires "nim >= 2.0.8"
-requires "https://github.com/crashappsec/con4m#5d58901a470c38ab0a242f0806d1ad08436a2fc7"
+requires "https://github.com/crashappsec/nimutils#c3b44fa9c913ccbace06fe1e5ce28b1f5cc7e22b"
 requires "https://github.com/viega/zippy == 0.10.7" # MIT
 
 # this allows us to get version externally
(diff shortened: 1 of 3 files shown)


miki725:
for some clarity, we dont be removing code from con4m or sending prs to that repo. instead once we start integrating with lib4m we will just drop the con4m repo dep altogether and will instead copy-over just the bits which will be required (if any)

viega:
100% agree with @miki725

PR #419, opened 2024-09-05
feat: adding support for CHALK_BYPASS environment variable

### LINKED ISSUE #224: Support a NOP run mode beyond virtual chalking

miki725:

## Issue

there is no way to disable/enable chalk when chalk wraps a `docker` command

fixes https://github.com/crashappsec/chalk/issues/224

## Description

this allows to completely skip chalk when wrapping `docker`. Whenever its enabled, chalk directly proxies the command to `docker` without invoking any of the chalk internal machinery such as loading configs/etc

## Testing

```
➜ make tests args="test_docker.py::test_docker_default_command --logs"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@
 - `_OP_CLOUD_SYS_VENDOR` key for reporting sys vendor
   file content used to identity cloud provider.
   ([#418](https://github.com/crashappsec/chalk/pull/418))
+- Supporting `CHALK_BYPASS` environment variable which
+  bypasses chalk altogether when chalk wraps `docker`.
+  ([#419](https://github.com/crashappsec/chalk/pull/419))
 
 ## 0.4.12
 
--- a/src/bypass.nim
+++ b/src/bypass.nim
@@ -1,0 +1,37 @@
+##
+## Copyright (c) 2024, Crash Override, Inc.
+##
+## This file is part of Chalk
+## (see https://crashoverride.com/docs/chalk)
+##
+
+import std/[os, strutils]
+import pkg/nimutils/[file]
+import "."/[util]
+
+proc bypassChalk*() =
+  ## Bypasses any chalk execution if possible and directly execs the underlying command.
+  ## Note this function explicitly writes to stderr for "logging" to avoid invoking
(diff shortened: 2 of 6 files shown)


miki725:
https://app.shortcut.com/crashoverride-1/story/837/support-a-nop-run-mode-beyond-virtual-chalking

ee7 (dismissed):
LGTM, assuming that e.g. @drraid is happy with the mechanism here.

ee7 (approved):

drraid (commented):
If I understand the code correctly, the default behavior of this is to still use chalk, and only bypass if the env var `CHALK_BYPASS`  is true (1, etc.), and I think this is the inversion of what we want: if someone has configured chalk to use bypass, it should default to bypass unless an env var specifies that it shouldn't. The recent request to implement this is for a scenario where the customer wants people to be able to opt-in by setting a variable, and also this opt-in approach prevents accidentally CRWDSTRIKEing their machines for replaced binaries. 

Maybe we want something like `PERFORM_WRAPPED_CHALKING`? Or maybe even something more explicit to identify the docker scenario? 

Additionally I'd like to do some analysis to figure out the code paths before this code is reached, the goal will be to minimize it as much as possible so that bugs elsewhere in the chalk code base can't cause a failure of our proxying of execution

PR #425, opened 2024-09-25
feat: adding pingttl util

miki725:

## Issue

get some metadata when imds ip endpoint cannot be reached

## Description

This will allow to get transient IP addresses along the NAT route so that we can fetch intermediate IPs from within docker containers which might not be accessible otherwise. For example it can be used to get the subnet gateway IP which is not otherwise available as within the docker container unless its running with host networking chalk can only get the docker network private IP address. This should be useful in figuring out where within the network container is running.

## TODO

- [x] this works outside of chalk however some linux headers cannot be resolved inside chalk
- [x] this depends on linux data-structs so this should be only compiled for linux, not mac
- [x] tests

## Tests

```
➜ make tests args="test_plugins.py::test_network --logs"
```

### DIFF
--- a/src/configs/base_keyspecs.c4m
+++ b/src/configs/base_keyspecs.c4m
@@ -5667,7 +5667,7 @@ keyspec _NETWORK_PARTIAL_TRACEROUTE_IPS {
     type:             dict[string, list[string]]
     standard:         true
     since:            "0.4.13"
-    shortdoc:         "Partial traceroute IPs to discover local network resources"
+    shortdoc:         "Partial traceroute IPs for run-time environment discovery"
     doc:              """
 Partial traceroutes to IPs as configured by `network.tracerroute_ips`.
 This allows to discover local network resources such as local
--- a/src/configs/base_keyspecs.c4m
+++ b/src/configs/base_keyspecs.c4m
@@ -5662,6 +5662,20 @@ AWS where imdsv2 is available.
 """
 }
 
+keyspec _NETWORK_PARTIAL_TRACEROUTE_IPS {
+    kind:             RunTimeHost
+    type:             dict[string, dict[string, string]]
+    standard:         true
+    since:            "0.4.13"
+    shortdoc:         "Partial traceroute IPs for configured TTLs"
+    doc:              """
+For all IPs to traceroute, intermediate IPs for given TTL values
+as configured by `network.traceroute_ips`.
+As this is not a complete traceroute all the way to the destination,
+it is useful to find local network configurations such as subnet gateway
+where chalk is running.
+"""
+}
 
 keyspec _CHALK_EXTERNAL_ACTION_AUDIT {
     kind: RunTimeHost
--- a/src/configs/base_keyspecs.c4m
+++ b/src/configs/base_keyspecs.c4m
@@ -5664,16 +5664,14 @@ AWS where imdsv2 is available.
 
 keyspec _NETWORK_PARTIAL_TRACEROUTE_IPS {
     kind:             RunTimeHost
-    type:             dict[string, dict[string, string]]
+    type:             dict[string, list[string]]
     standard:         true
     since:            "0.4.13"
-    shortdoc:         "Partial traceroute IPs for configured TTLs"
+    shortdoc:         "Partial traceroute IPs to discover local network resources"
     doc:              """
-For all IPs to traceroute, intermediate IPs for given TTL values
-as configured by `network.traceroute_ips`.
-As this is not a complete traceroute all the way to the destination,
-it is useful to find local network configurations such as subnet gateway
-where chalk is running.
+Partial traceroutes to IPs as configured by `network.tracerroute_ips`.
+This allows to discover local network resources such as local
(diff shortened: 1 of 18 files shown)


src/configs/base_keyspecs.c4m
viega, line 1: Seems like it'd be easy to have a configurable hop limit from the start instead of a key like this whose purpose isn't going to be particularly evident from the name, no??  I'm OK with it, but if we're going to want to ditch the name when we add the configurability, it maybe is worth the time to do it now, because it really should be minor.

src/configs/base_plugins.c4m
```diff
@@ -231,6 +231,15 @@ Captures metadata from cloud instances.
 """
 }
 
+plugin network {
```
viega, line 234: ALSO, it JUST occurs to me, that you could already use the proc plugin to get the network routing table, which might help here! That def doesn't require any code changes for this PR, but it might make sense to expose such things as more high-level construct in this plugin over time.

src/pingttl.nim
```diff
@@ -0,0 +1,270 @@
...
+          raiseOSError(lastError)
+        if int32(offender.sa_family) != posix.AF_INET:
+          trace("pingttl: icmp error from unsupported offender source address family " & $offender.sa_family)
+          raiseOSError(lastError)
+        # TODO how to get the sock len directly from the sockaddr structure?
+        # var
+        #   address: IpAddress
+        #   port:    Port
+        # fromSockAddr(cast[ptr Sockaddr_in](offender)[], ???, address, port)
+        let address = cast[ptr Sockaddr_in](offender)
```
viega, line 199: My recollection is that Nim has a sizeof operator like C, so you can just alloc something of the right size. You also can just always allocate to the size of the `sockaddr_in6` struct, and pass it; the over-allocation isn't an issue when it's an ipv4 connection (in fact, that's essentially how I do it in libcon4m for the IP address type, I always allocate `sizeof(struct sockaddr_in6)`).  But this is all easy to do from C too, if needed.

viega (approved):
It's fine to go out like this, your choice on whether to take it a little farther based on my notes for sure :)

src/pingttl.nim
viega: The big Nim gotcha BTW is that you have to worry about the Nim lifetime for anything that's getting passed to C. Probably won't be an issue here, but there are odd scenarios where the Nim semantics won't be obvious.

src/configs/base_keyspecs.c4m
```diff
@@ -5662,6 +5662,20 @@ AWS where imdsv2 is available.
 """
 }
 
+keyspec _NETWORK_PARTIAL_TRACEROUTE_IPS {
+    kind:             RunTimeHost
+    type:             dict[string, dict[string, string]]
+    standard:         true
+    since:            "0.4.13"
+    shortdoc:         "Partial traceroute IPs for configured TTLs"
```
drraid, line 5670: I'd suggest that this shortdoc should include the "why" or "what it means" more than the technical "what", maybe something like:
"Partial tracerouting for run-time environment discovery" ? I'm not sure what it should be, but I'd prefer any description to be more around its rationale for existing

src/configs/base_plugins.c4m
miki725: in docker container, the proc will have routing table for the docker gateway so prolly something like `172.17.0.1` whereas the thing we actually care about is the next hop over which wont be anywhere in the proc as far as I know

also we already do collect all interfaces from the proc as another key

https://github.com/crashappsec/chalk/blob/697624bac82050ca847c4c779ae36e7ec5dafa6d/src/configs/base_keyspecs.c4m#L4659-L4680

src/configs/base_keyspecs.c4m
miki725: sure. the idea was to allow to discover intermediate ips along the route. for example in AWS it will be common for the egress NAT to be 3 hops away (container, gateway, NAT) and some people might not care about closer IPs but might want to know some IP a little ways off but I agree its not intuitive. will change to just specify number of hops to traceroute per destination IP

miki725: updated. now the report will look something like:

```
    "_NETWORK_PARTIAL_TRACEROUTE_IPS": {
      "1.1.1.1": [
        "172.25.0.1",
        "10.10.20.1"
      ]
    },
```

miki725:
https://app.shortcut.com/crashoverride-1/story/2349/chalk-add-simple-traceroute-for-docker-containers-to-report-host-ip

src/configs/base_keyspecs.c4m
miki725: thanks. updated

src/pingttl.nim
miki725: thanks. directly passing sizeof ipv4 worked. didnt go for the ipv6 as above error handling already asserts its ipv4 error as otherwise well need to handle ipv6 errors differently. did add a comment so when we need to add ipv6 support it shouldnt be too bad. just diff error codes and sock options

viega (approved):

PR #431, opened 2024-10-13
feat: adding docker registry mirrors support

miki725:

## Issue

adds support for mirrors in both docker daemon config as well as buildx node configs

## Description

Whenever mirror is configured, docker pulls images from the mirror unless there is an error in which case it falls back to the upstream registry. Chalk should match that behavior so that we can both interact with the registries in the same manner as docker as well as in the future be able to report metadata about registry use more accurately.

## Testing

```
➜ make tests args="test_docker.py::test_base_registry --logs"
```

**note** this is PR to upstream `registry` branch

### DIFF
--- a/src/docker/ids.nim
+++ b/src/docker/ids.nim
@@ -182,12 +182,6 @@ proc withDigest*(items: seq[DockerImage], digest: string): seq[DockerImage] =
   for i in items:
     result.add(i.withDigest(digest))
 
-proc registry*(self: DockerImage): string =
-  return self.repo.split('/', maxsplit = 1)[0]
-
-proc domain*(self: DockerImage): string =
-  return self.registry.split(':', maxsplit = 1)[0]
-
 proc isFullyQualified(self: DockerImage): bool =
   ## determine if the docker image is a fully qualified image name
   ## as in if the image should be pulled/pushed to default docker registry (docker hub)
@@ -213,7 +207,7 @@ proc isFullyQualified(self: DockerImage): bool =
     maybeRegistry.toLower() != maybeRegistry
   )
 
-proc qualify*(self: DockerImage): DockerImage =
+proc qualify(self: DockerImage): DockerImage =
   ## fully qualify image name with the full registry domain
   ## note qualified name does not specify any scheme like http or https
   ## it simply specifies complete image reference in the registry
@@ -230,7 +224,22 @@ proc qualify*(self: DockerImage): DockerImage =
     self.digest,
   )
 
-proc normalize*(self: DockerImage): DockerImage =
+proc withRegistry*(self: DockerImage, registry: string): DockerImage =
+  if registry == "":
+    return self
+  # TODO handle paths in registry
+  # parseUri doesnt parse uri without any scheme
+  let qualified = self.qualify()
+  var uri       = parseUri("https://" & qualified.repo)
+  uri.hostname  = registry
+  let repo = ($uri).removePrefix("https://")
+  result = (
+    repo,
+    self.tag,
+    self.digest,
+  )
+
+proc normalize(self: DockerImage): DockerImage =
   ## normalize qualified registry domain
   ## normalization maps some hardcoded registry domains
   ## to their standard API domains
@@ -239,16 +248,12 @@ proc normalize*(self: DockerImage): DockerImage =
     "docker.io":       DEFAULT_REGISTRY,
     "index.docker.io": DEFAULT_REGISTRY,
   }.toTable()
-  let qualified = self.qualify()
-  # parseUri doesnt parse uri without any scheme
-  var uri       = parseUri("https://" & qualified.repo)
-  uri.hostname  = registryMapping.getOrDefault(uri.hostname, uri.hostname)
-  let repo = ($uri).removePrefix("https://")
-  result = (
-    repo,
-    self.tag,
-    self.digest,
-  )
+  let
+    qualified = self.qualify()
+    registry  = qualified.repo.split('/', maxsplit = 1)[0]
+  if registry in registryMapping:
+    return qualified.withRegistry(registryMapping[registry])
+  return qualified
 
 proc uri*(self: DockerImage, scheme = "", path = "", healthcheck = false): Uri =
   ## generate working URI for the registry API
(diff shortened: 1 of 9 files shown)


src/docker/ids.nim
```diff
@@ -213,72 +227,102 @@ proc isFullyQualified(self: DockerImage): bool =
...
   ## note this only supports v2 registries hence hardcodes v2 suffix
   ## also this doesnt account for any insecure registry configs
   let normalized = self.normalize()
-  var uri = parseUri("https://" & normalized.repo)
-  if healthcheck:
-    uri.path = "/v2/"
-  else:
-    uri.path = "/v2" & uri.path
+  var uri        = parseUri("https://" & normalized.repo)
+  let uriPath    = uri.path
```
mbaltrusitis, line 275: does this end up being a value or ref?

src/docker/registry.nim
```diff
@@ -39,6 +51,14 @@ const
     "application/vnd.oci.image.config.v1+json": DockerManifestType.config,
   }.toTable()
 
+iterator uses(use: RegistryUse): RegistryUse =
+  ## which uses lookups are applicable for the registry use
+  ## ReadOnly use can only be used for reads
+  ## however ReadWrite use can be used for both
+  yield use
+  if use == RegistryUse.ReadWrite:
```
mbaltrusitis, line 59: consider renaming to something like "RegistryCapability" or "RegistryPermission". The callsites below are hard to read / reason about

src/docker/registry.nim
```diff
@@ -60,25 +91,38 @@ iterator iterDaemonRegistryConfigs(self: DockerImage): RegistryConfig =
...
-  template yieldInsecure() =
-    trace("docker: " & self.registry & " will attempt TLS without verifying server cert")
-    yield RegistryConfig(scheme: "https://", verifyMode: CVerifyNone)
-    yield RegistryConfig(scheme: "http://", verifyMode: CVerifyNone)
-
   for i in getDockerInfoSubList("insecure registries:"):
     if self.registry == i or self.domain == i:
       trace("docker: " & i & " is configured as an insecure registry in docker daemon")
-      yieldInsecure()
+      trace("docker: " & self.registry & " will attempt TLS without verifying server cert")
```
mbaltrusitis, line 110: does nim have something similar to __module__ or __file__ so we can avoid manually putting the `docker:` prefix in?

src/docker/registry.nim
```diff
@@ -269,20 +448,28 @@ proc manifestHead*(image: DockerImage): DockerDigestedJson =
...
       path       = "/manifests/" & image.imageRef,
       accept     = accept,
     )
   return newDockerDigestedJson(response.body(), image.imageRef, accept, kind)
 
-proc layerGet*(image: DockerImage, accept: string): DockerDigestedJson =
+proc layerGet*(image:  DockerImage,
+               accept: string,
+               use =   RegistryUse.ReadOnly,
+               ): DockerDigestedJson =
```
mbaltrusitis, line 468: nit: closing paren and return type should be raised to next line to be consistent with the rest of the PR.

tests/functional/conf.py
```diff
@@ -45,6 +45,7 @@
 REGISTRY = f"{os.environ.get('IP') or 'localhost'}:5044"
 REGISTRY_TLS = f"{os.environ.get('IP') or 'localhost'}:5045"
 REGISTRY_TLS_INSECURE = f"{os.environ.get('IP') or 'localhost'}:5046"
+REGISTRY_PROXY = f"{os.environ.get('IP') or 'localhost'}:5047"
```
mbaltrusitis, line 48: is there a reason not to pass the default value to the `.get(key, default_val)` method?

REGISTRY_PROXY = f"{os.environ.get('IP', 'localhost'}:5047"

src/docker/registry.nim
miki725: this log function is from con4m, not nim builtin. in there we have to provide full log string all the time

src/docker/registry.nim
miki725: its not a registry permission or capability question. some "mirrors" allow you to manually push to them explicitly so the question here is in what context that registry being used, not an absolute capability/permission registry has

for example if you have `1.2.3.4` configured as registry mirror. these are all valid:

```
docker pull alpine # will attempt 1.2.3.4 first
docker pull 1.2.3.4/alpine
docker push alpine # will NOT attempt 1.2.3.4
docker push 1.2.3.4/alpine
```

src/docker/ids.nim
miki725:
```nim
Uri = object
  scheme*, username*, password*: string
  hostname*, port*, path*, query*, anchor*: string
  opaque*: bool
  isIpv6*: bool
```

as `path` is `string`, not `ref string`, it will be value

tests/functional/conf.py
miki725: that env var is defaulted in compose file to empty string

https://github.com/crashappsec/chalk/blob/a988c7b4a85608fe5a000cce999827dba0cd4d8e/docker-compose.yml#L159-L159

when not using `make` to invoke `docker compose` which sets `IP`:

https://github.com/crashappsec/chalk/blob/a988c7b4a85608fe5a000cce999827dba0cd4d8e/Makefile#L103-L103

its value is going to be empty string hence `.get()` will return empty string which wont be correct

with `or` it gets correctly defaulted:

```
➜ docker compose run --rm --no-deps --entrypoint=python tests -c 'import conf; print(conf.REGISTRY_PROXY)'
localhost:5047
```

src/docker/registry.nim
miki725: the pattern we seem to use is:

sing-line proc signature:

```nim
proc foo(a: string): int=
```

when there are a lot of params we break it up into multiple lines:

```nim
proc foo(a: string,
         b: string,
         ): int =
```

eventually will be nice to get proper autoformatter involved too. at some nim switched recommeded format from hanging to:

```nim
proc foo(
  a: string,
  b: string,
): int =
```

I actually prefer that as it matches `black` more

mbaltrusitis: i like the last option best as well. keeps the interesting bits away from too much indentation.

src/docker/registry.nim
miki725: yep. we can rename to `useCase` though. that sounds good and is accurate.

although again, wont do in this pr. will collate all the comments in a follow-up PR

mbaltrusitis (approved):

PR #433, opened 2024-10-15
refactor: extracting chalk mark from registry

miki725:

## Description

chalk can extract chalk mark from either:

* in-toto attestation layer
* last layer directly from registry

this will be necessary for base image inspection as well as dropping cosign dep in the future

also now we can directly grab provenance/sbom from the registry

## Testing

run `test_docker.py` tests. all existing functionality should still work as-is

### DIFF
--- a/src/docker/extract.nim
+++ b/src/docker/extract.nim
@@ -208,6 +208,7 @@ proc extractImage*(self: ChalkObj) =
   if self.imageDigest != "":
     try:
       self.extractFrom(self.extractMarkFromSigStore(), self.getImageName())
+      self.signed = true
       info("docker: " & self.getImageName() & ": chalk mark successfully extracted from in-toto attestation from registry")
       return
     except:
@@ -224,7 +225,7 @@ proc extractImage*(self: ChalkObj) =
 
   if self.canVerifyBySigStore():
     try:
-      self.extractFrom(self.extractMarkFromSigStore(), self.getImageName())
+      self.extractFrom(self.extractMarkFromSigStoreCosign(), self.getImageName())
       self.signed = true
       info("docker: " & self.getImageName() & ": chalk mark successfully extracted from attestation.")
       return
--- a/src/docker/extract.nim
+++ b/src/docker/extract.nim
@@ -215,14 +215,6 @@ proc extractImage*(self: ChalkObj) =
       trace("docker: " & self.getImageName() & ": could not extract chalk mark " &
             "via in-toto attestation due to: " & getCurrentExceptionMsg())
 
-    try:
-      self.extractFrom(self.extractMarkFromLayer(), self.getImageName())
-      info("docker: " & self.getImageName() & ": chalk mark successfully extracted from layer from registry")
-      return
-    except:
-      trace("docker: " & self.getImageName() & ": could not extract chalk mark " &
-            "via layer in registry due to: " & getCurrentExceptionMsg())
-
   if self.canVerifyBySigStore():
     try:
       self.extractFrom(self.extractMarkFromSigStoreCosign(), self.getImageName())
@@ -233,6 +225,15 @@ proc extractImage*(self: ChalkObj) =
       self.noCosign = true
       trace("docker: " & self.getImageName & ": could not extract chalk mark " &
            "via attestation due to: " & getCurrentExceptionMsg())
+
+  if self.imageDigest != "":
+    try:
+      self.extractFrom(self.extractMarkFromLayer(), self.getImageName())
+      info("docker: " & self.getImageName() & ": chalk mark successfully extracted from layer from registry")
+      return
+    except:
+      trace("docker: " & self.getImageName() & ": could not extract chalk mark " &
+            "via layer in registry due to: " & getCurrentExceptionMsg())
 
   self.extractFrom(self.extractImageMark(), self.getImageName())
   info("docker: " & self.getImageName() & ": chalk mark successfuly extracted from image.")
--- a/src/docker/extract.nim
+++ b/src/docker/extract.nim
@@ -112,10 +112,10 @@ proc extractMarkFromLayer(self: ChalkObj): string =
           "Last layer for " & $spec & " does not contain /chalk.json",
         )
       let lastLayer = manifest.layers[^1]
-      return lastLayer.nameRef.layerGetFSFileStream(
+      return lastLayer.nameRef.layerGetFSFileString(
         name   = "chalk.json",
         accept = manifest.mediaType,
-      ).readAll()
+      )
     except:
       err = getCurrentExceptionMsg()
       trace("docker: could not extract chalk mark from registry: " & err)
--- a/src/docker/extract.nim
+++ b/src/docker/extract.nim
@@ -9,7 +9,7 @@
 
 import std/[base64]
 import ".."/[attestation/utils, config, chalkjson, util]
-import "."/[exe, inspect, ids]
+import "."/[exe, inspect, ids, manifest, registry]
 
 proc hasChalkLayer(self: ChalkObj): bool =
   result = false
@@ -17,6 +17,17 @@ proc hasChalkLayer(self: ChalkObj): bool =
   for layer in layers:
     if "COPY " in layer and " /chalk.json" in layer:
       return true
+
+proc hasChalkLayer(self: DockerManifest): bool =
+  if self.kind != DockerManifestType.image:
+    raise newException(ValueError, "Can only query chalk marks from manifest image. Given " & $self.kind)
+  let history = self.config.json{"history"}
+  if history == nil or history.kind != JArray or len(history) == 0:
+    raise newException(ValueError, $self.name & " image config does not have valid history")
+  let
+    lastStep  = history[^1]
+    createdBy = lastStep{"created_by"}.getStr()
+  return "COPY" in createdBy and "/chalk.json" in createdBy
 
 proc extractContainerMark(containerId: string): string =
   let (markStream, markTmp) = getNewTempFile(suffix = "chalk.json")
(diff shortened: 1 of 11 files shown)


miki725:
https://app.shortcut.com/crashoverride-1/story/914/chalk-fetch-chalkmarks-from-registry-attestestation-if-present

miki725:
https://app.shortcut.com/crashoverride-1/story/915/chalk-fetch-chalkmarks-from-registry-layer-if-present

miki725:
https://app.shortcut.com/crashoverride-1/story/1760/chalk-pull-sbom-provenance-via-registry-api-wrappers-vs-buildx-imagetools

src/docker/manifest.nim
```diff
@@ -310,34 +280,33 @@ proc fetchManifest*(name: DockerImage, otherNames: seq[DockerImage] = @[]): Dock
...
-  var manifest = fetchManifest(name)
+proc fetchOnlyImageManifest*(name: DockerImage, fetchConfig = true): DockerManifest =
+  var manifest = fetchManifest(name, fetchConfig = fetchConfig)
   if manifest.kind == DockerManifestType.list:
     let manifests = manifest.findAllPlatformsManifests()
     if len(manifests) == 1:
       manifest = manifests[0]
-      manifest.fetch()
     else:
       raise newException(KeyError, "There are multiple platform images for: " & $name)
```
mbaltrusitis, line 300: Would adding a platform arg here, defaulting to the host platform, alleviate the multiple platform image exception?

src/docker/manifest.nim
```diff
@@ -351,3 +320,94 @@ proc fetchImageManifest*(name: DockerImage,
...
+      return i.fetch()
+  raise newException(KeyError, "Could not find sibling image of reference type: " & reference)
+
+proc findInTotoLayer(self: DockerManifest, predicate: string): DockerManifest =
+  if self.kind != DockerManifestType.image:
+    raise newException(ValueError, "Can only lookup layers in image manifest")
+  for l in self.layers:
+    if l.annotations == nil:
+      continue
+    if l.annotations{"in-toto.io/predicate-type"}.getStr().toLower().startsWith(predicate.toLower()):
```
mbaltrusitis, line 345: `333,334,345`: util function for case-insensitive string comparisons?

src/docker/manifest.nim
```diff
@@ -351,3 +320,94 @@ proc fetchImageManifest*(name: DockerImage,
...
+      # for multi-platform we have to filter on the platform
+      # note index only supports a few keys so have to manually
+      # take SLSA key
+      return requestManifestJson(
+        name,
+        flags    = @["--format", "{{json (index .Provenance \"" & $platform & "\")}}"],
+        fallback = false,
+      ).json{"SLSA"}
+
+proc fetchSBOM*(name: DockerImage, platform: DockerPlatform): JsonNode =
```
mbaltrusitis, line 382: nit: `fetchSBOM` -> `fetchSbom`. Would align with Nim's std lib style (e.g. `JsonNode`)

src/docker/registry.nim
```diff
@@ -369,12 +370,19 @@ iterator getConfigs(self: DockerImage, use: RegistryUse): RegistryConfig =
     if not isBuildx:
       buildx()
 
+var jsonCache = initTable[
+  (DockerImage, HttpMethod, string, RegistryUse),
+  (string, Response)
+]()
```
mbaltrusitis, line 376: to avoid global mutation, could this be its own module with set/get API to keep access to this state more uniform?

src/docker/registry.nim
```diff
@@ -462,16 +471,46 @@ proc manifestGet*(image:  DockerImage,
...
       path       = "/blobs/" & image.imageRef,
       accept     = accept,
     )
-  return newDockerDigestedJson(response.body(), image.imageRef, accept, kind)
+  return response.body()
+
+proc layerGetJson*(image:  DockerImage,
+                   accept: string,
+                   use =   RegistryUse.ReadOnly,
+                   ): DigestedJson =
```
mbaltrusitis, line 490:
```suggestion
proc layerGetJson*(
    image:  DockerImage,
    accept: string,
    use =   RegistryUse.ReadOnly,
): DigestedJson =
```

mbaltrusitis: inline with our style discussion yesterday

src/docker/registry.nim
```diff
@@ -462,16 +471,46 @@ proc manifestGet*(image:  DockerImage,
...
+      accept = accept,
+    ),
+    digest = image.imageRef,
+  )
+
+proc layerGetFSFileString*(image:  DockerImage,
+                           name:   string,
+                           accept: string,
+                           use =   RegistryUse.ReadOnly,
+                           ): string =
```
mbaltrusitis, line 503:
```suggestion
proc layerGetFSFileString*(
    image:  DockerImage,
    name:   string,
    accept: string,
    use =   RegistryUse.ReadOnly,
): string =
```

src/docker/manifest.nim
```diff
@@ -351,3 +320,94 @@ proc fetchImageManifest*(name: DockerImage,
...
+  # https://docs.docker.com/reference/cli/docker/buildx/imagetools/inspect/
+  try:
+    trace("docker: looking up SBOM for: " & $name)
+    let layer = name.fetchImageManifest(platform).findSibling().findInTotoLayer("https://spdx.dev/Document")
```
mbaltrusitis, line 386: what do you think about having these "pipeline-like" calls with 2 or more transformations being brought onto their own lines to ease grokking how something is being mutated? Something like:

```suggestion
let layer = name.fetchImageManifest(platform)
              .findSibling()
              .findInTotoLayer("https://spdx.dev/Document")
```

src/docker/extract.nim
```diff
@@ -88,7 +99,49 @@ proc extractImageMark(self: ChalkObj): string =
...
+      if not manifest.hasChalkLayer():
+        raise newException(
+          ValueError,
+          "Last layer for " & $spec & " does not contain /chalk.json",
+        )
+      let lastLayer = manifest.layers[^1]
+      return lastLayer.nameRef.layerGetFSFileString(
+        name   = "chalk.json",
+        accept = manifest.mediaType,
+      )
```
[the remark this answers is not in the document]
miki725: I prefer  pattern to exit early on error so that the rest of the function doesnt need to be indented as much

```
for i in foo():
  if i.bad():
    continue
  i.good()
```

I especially prefer this when the `good()` logic is more involved where its a dozens of lines. having everything one more indent I think doesnt help, especially when splitting buffers where the left real-estate is more precious

src/docker/extract.nim
```diff
@@ -88,7 +99,49 @@ proc extractImageMark(self: ChalkObj): string =
...
+    except:
+      err = getCurrentExceptionMsg()
+      trace("docker: could not extract chalk mark from registry: " & err)
+      continue
+  raise newException(ValueError, err)
+
+proc extractMarkFromInToto(self: ChalkObj, json: JsonNode): string =
```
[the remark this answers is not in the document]
miki725: yeah its a bit weird name of the format

src/docker/manifest.nim
```diff
@@ -161,12 +164,14 @@ proc setImageLayers(self: DockerManifest, data: DigestedJson) =
...
       digest:        layer{"digest"}.getStr(),
       size:          layer{"size"}.getInt(),
+      annotations:   layer{"annotations"},
     ))
 
-proc fetch(self: DockerManifest) =
+proc fetch(self: DockerManifest, fetchConfig = true): DockerManifest {.discardable.} =
+  result = self
   if self.isFetched:
     return
```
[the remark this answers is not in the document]
miki725: strictly speaking not necessary but sure. I think Ive done both:

```
➜ rg 'return result' | wc -l
22
```

src/docker/manifest.nim
miki725: no

this is used in https://github.com/crashappsec/chalk/blob/84d24ad22bbbfa3ecd336e7bdf412938ed98398d/src/docker/platform.nim#L107 which is inside https://github.com/crashappsec/chalk/blob/84d24ad22bbbfa3ecd336e7bdf412938ed98398d/src/docker/platform.nim#L75-L76 which is required for https://github.com/crashappsec/chalk/blob/84d24ad22bbbfa3ecd336e7bdf412938ed98398d/src/docker/platform.nim#L265-L269

there are a few fallbacks in order to correctly determine correct platform of the build. it does not always match host platform.

for example:

```dockerfile
FROM foo:arm64
CMD foo
```

this will build arm64 image even if the host platform is amd64. what `foo:arm64` is has 2 possibilities though:

* its a manifest list which has a single platform image inside of it. for example this will happen when enabling provenance but not doing multi-platform build. it will yield manifest list with a single build in it where the second reference will be attestation itself
* its not a manifest list but an image manifest directly which will have platform specified

code for that: https://github.com/crashappsec/chalk/blob/84d24ad22bbbfa3ecd336e7bdf412938ed98398d/src/docker/platform.nim#L105-L114

so in this case we dont care the platform itself, just care that there is a single platform in the manifest list

src/docker/manifest.nim
miki725: 🤷 . I like using explicit `==` for comparisons as most editors highlight `==` differently vs utility will have to be something like `a.foo(b)` which wont use `==` unless we add string wrapper around it like `Caseless(a) == b`, it can convert inside `==` implicitly but that can be hard to follow as well

as its all one-line, prolly no big deal but no strong feeling either way

src/docker/manifest.nim
miki725: SBOM is in itself abbreviation. most cases we use `SBOM`:

```
➜ rg 'SBOM' | wc -l
73
➜ rg 'Sbom' | wc -l
2
```

we should prolly change those 2 to match `SBOM` everywhere else

src/docker/registry.nim
miki725: 🤷 for now its meant to be transparent on the caller when using registry SDK, stuff just gets cached transparently. we use this pattern in a bunch of places. eventually we can add util to manage cache however dont think it will win much in terms of consiseness

src/docker/registry.nim
miki725: most of chalk uses hanging indents for function signatures now. when we switch to auto-formatter we can switch to something else then. for now we can keep as-is?

src/docker/manifest.nim
miki725: this is invalid syntax in nim I think. I think `.` needs to stay on the previous line or something like that when last time I tried to use that style of chained functions

src/docker/manifest.nim
mbaltrusitis: clear. thank you.

mbaltrusitis (dismissed):

mbaltrusitis (approved):

src/docker/manifest.nim
miki725: actually its a valid syntax. changed in https://github.com/crashappsec/chalk/pull/465

src/docker/manifest.nim
miki725: changed in https://github.com/crashappsec/chalk/pull/465

PR #434, opened 2024-10-15
feat: heartbeat minimal report template

miki725:

## Issue

Default heartbeat report is too large as it uses the default report which is quite heavy for simple liveliness checks

## Description

This adds minimal report template for heartbeats which simply reports exec id (to tie to exec report) as well as other basic metadata such as metadata id, operation, timestamp, etc.

Will be enabling it by default specifically for co config later.

## Testing

```
➜ make tests args="test_docker.py::test_docker_heartbeat --logs --pdb"
```

### DIFF
--- a/src/configs/base_report_templates.c4m
+++ b/src/configs/base_report_templates.c4m
@@ -332,6 +332,7 @@ report and subtract from it.
   key._SIGNATURES.use                         = true
   key._INVALID_SIGNATURE.use                  = true
   key._ACTION_ID.use                          = true
+  key._EXEC_ID.use                            = true
   key._ARGV.use                               = true
   key._ENV.use                                = true
   key._TENANT_ID.use                          = true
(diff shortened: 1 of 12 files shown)


miki725:
https://app.shortcut.com/crashoverride-1/story/2760/chalk-enable-heartbeats-by-default

src/configs/base_report_templates.c4m
```diff
@@ -1907,6 +1912,26 @@ report_template usage {
...
+"""
+  ~key.METADATA_ID.use                        = true
+  ~key.CHALK_ID.use                           = true
+  ~key._CHALKS.use                            = true
+  ~key._OPERATION.use                         = true
+  ~key._TIMESTAMP.use                         = true
+  ~key._DATETIME.use                          = true
+  ~key._ACTION_ID.use                         = true
+  ~key._EXEC_ID.use                           = true
+  ~key._OP_ARGV.use                           = true
```
miki725, line 1932: not sure if this key is strictly necessary although seems like a useful sanity check when viewing the report that its for the correct execution? as otherwise `_EXEC_ID` is a random string so not super friendly...

thoughts whether we should keep it in or remove it from heartbeat reports?

nettrino (approved):

viega (approved):

nettrino (approved):

PR #443, opened 2024-11-07
fix: safeRequest memory leak

miki725:

## Issue

chalk running in as daemon with heartbeats had ever-increasing memory footprint

## Description

Each heartbeat report, it would send HTTP request to chalk api. That was done via `safeRequest` from nimutils which would initialize `SslContext`. That context included all root CA certs/etc and would use ~1Mb on average. At the end of the request, it wasnt correctly discarded as nim would GC its reference however it would call into openssl functions to clear all the inner data-structures. Now we explicitly destroy the context after each request to ensure its memory is correctly cleared.

for more:

https://github.com/crashappsec/nimutils/pull/81

## Tests

Run long running process with heartbeats enabled and check its memory utilization.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,11 @@
   chalk-mark level. For example `SBOM` key with equivalent
   content was duplicated twice.
   ([#440](https://github.com/crashappsec/chalk/pull/440))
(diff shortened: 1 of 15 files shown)


viega (approved):
Excellent thanks, Miro.

PR #454, opened 2024-12-02
fix: DOCKERFILE_PATH_WITHIN_VCTL cannot have :stdin: as value

mbaltrusitis:
If the contents of a Dockerfile is passed via stdin (e.g. FROM alpine | chalk build) the key reporting the relative path should be dropped.

## Description

The predicates that would trigger the dropping of `DOCKERFILE_PATH_WITHIN_VCTL` were insufficient in the case where a Dockerfile's contents was provided via a pipe (e.g., `FROM alpine | chalk docker build -f - .`). This resulted in the reporting of values like `:stdin:` or `some/path/:stdin:`.

## Testing

make tests args="test_docker.py::test_piped_dockerfile_has_no_relative_path --logs

### DIFF
--- a/src/docker/build.nim
+++ b/src/docker/build.nim
@@ -7,7 +7,6 @@
 
 import std/json
 import ".."/[
-  chalk_common,
   chalkjson,
   collect,
   config,
--- a/src/docker/build.nim
+++ b/src/docker/build.nim
@@ -7,6 +7,7 @@
 
 import std/json
 import ".."/[
+  chalk_common,
   chalkjson,
   collect,
   config,
@@ -51,7 +52,7 @@ proc processGitContext(ctx: DockerInvocation) =
     raise
 
 proc processDockerFile(ctx: DockerInvocation) =
-  if ctx.dockerFileLoc == ":stdin:":
+  if ctx.dockerFileLoc == stdinIndicator:
     let input           = stdin.readAll()
     ctx.inDockerFile  = input
     ctx.originalStdIn = input
--- a/src/docker/cmdline.nim
+++ b/src/docker/cmdline.nim
@@ -11,7 +11,7 @@
 ## is in configs/dockercmd.c4m), so we really just need to look at
 ## the command and flag info returned.
 
-import ".."/[config]
+import ".."/[chalk_common, config]
 import "."/[ids]
 
 proc extractBuildx(ctx: DockerInvocation) =
@@ -71,7 +71,7 @@ proc extractDockerFile(ctx: DockerInvocation) =
     let files = unpack[seq[string]](ctx.processedFlags["file"].getValue())
     ctx.foundFileArg = files[0]
     if ctx.foundFileArg == "-":
-      ctx.dockerFileLoc = ":stdin:"
+      ctx.dockerFileLoc = stdinIndicator
     else:
       ctx.dockerFileLoc = resolvePath(ctx.foundFileArg)
 
--- a/src/docker/cmdline.nim
+++ b/src/docker/cmdline.nim
@@ -11,7 +11,7 @@
 ## is in configs/dockercmd.c4m), so we really just need to look at
 ## the command and flag info returned.
 
-import ".."/[chalk_common, config]
+import ".."/[config]
 import "."/[ids]
 
 proc extractBuildx(ctx: DockerInvocation) =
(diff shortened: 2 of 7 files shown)


tests/functional/test_docker.py
```diff
@@ -1600,3 +1600,10 @@ def test_version_bare(chalk_default: Chalk):
         params=["version"],
     )
     assert run
+
+
+def test_piped_dockerfile_has_no_relative_path(chalk: Chalk, random_hex: str):
+    tag_base = f"{REGISTRY}/nonchalked_{random_hex}"
+    tag = f"{tag_base}:latest"
+    _, result = chalk.docker_build(content="FROM alpine", tag=tag)
+    assert "DOCKERFILE_PATH_WITHIN_VCTL" not in result.mark
```
miki725, line 1609: this is fairly expensive test as it needs to do docker build, then chalked docker build just to assert a single field in the report. we already have other tests which pass dockerfile via stdin. can we adjust them to do this assertion as well. prolly will be something like:

```python
assert build.mark.has(
  ...
  DOCKERFILE_PATH_WITHIN_VCTL=MISSING,
)
```

src/util.nim
```diff
@@ -628,10 +628,10 @@ proc getOrDefault*[T](self: openArray[T], i: int, default: T): T =
 
 proc getRelativePathBetween*(fromPath: string, toPath: string) : string =
   ## Given the `fromPath`, usually the project root, return the relative
-  ## path of the file's `toPath`. Return nothing if its outside the project root
-  ## or if `toPath` is an empty string.
+  ## path of the file's `toPath`. Return nothing if its outside the project root,
+  ## if `toPath` is an empty string or, if Dockerfile contents was passed via stdin.
   result = toPath.relativePath(fromPath)
-  if result.startsWith("..") or result == "":
+  if result.startsWith("..") or result == "" or result.endsWith(":stdin:"):
```
miki725, line 634: 👍 

should we make the `:stdin:` string a constant somewhere so that its referred in the same way everywhere?

mbaltrusitis: good point

tests/functional/test_docker.py
mbaltrusitis: moved assertion under `test_base_images`

src/docker/build.nim
```diff
@@ -7,6 +7,7 @@
 
 import std/json
 import ".."/[
+  chalk_common,
```
miki725, line 10: no need to import `chalk_common`. we actually dont import it anywhere, except maybe a few places due to:

https://github.com/crashappsec/chalk/blob/73b7ef84c5da75d3b40fc00ffd7d6ee42c30c066/src/config.nim#L10-L11

which exports everything from `run_management` which in turn:

https://github.com/crashappsec/chalk/blob/73b7ef84c5da75d3b40fc00ffd7d6ee42c30c066/src/run_management.nim#L13-L14

therefore most modules in chalk simply import `config` and that automatically imports all the base stuff

src/docker/build.nim
```diff
@@ -51,8 +52,8 @@ proc processGitContext(ctx: DockerInvocation) =
     raise
 
 proc processDockerFile(ctx: DockerInvocation) =
-  if ctx.dockerFileLoc == ":stdin:":
-    let input           = stdin.readAll()
+  if ctx.dockerFileLoc == chalk_common.stdin:
```
miki725, line 55:
```suggestion
  if ctx.dockerFileLoc == stdin
```

you only need the specify module name whenever there is a conflict with another module which already has `stdin` defined

speaking of that as stdlib has `stdin` lets rename to something like `stdinIndicator` or something like that to indicate its not a file-like object but an indicator that something is stdin

mbaltrusitis: I had only included it where there was said conflict. I'll swap the name around.

src/docker/build.nim
miki725: config is already imported below. lets remove

```suggestion
```

src/docker/cmdline.nim
```diff
@@ -11,7 +11,7 @@
 ## is in configs/dockercmd.c4m), so we really just need to look at
 ## the command and flag info returned.
 
-import ".."/[config]
+import ".."/[chalk_common, config]
```
miki725, line 14:
```suggestion
import ".."/[config]
```

miki725 (approved):

PR #458, opened 2024-12-10
fix: correct git report datetimes

miki725:

## Issue

In some cases commit dates show as in the future.

## Description

When git reports its timestamp, its localized with the provided offset and as chalk wasnt accounting for that, in some cases, commit timestamps would show up as future timestamps which is very confusing

Now all timestamps are normalized with the provided offset.

In addition all date reporting was changed to use consistent output format of ISO-8601.

## Testing

```
➜ make tests args="test_git.py -x --logs --pdb"
```


src/plugins/system.nim
```diff
@@ -264,10 +256,10 @@ proc sysGetChalkTimeHostInfo*(self: Plugin): ChalkDict {.cdecl.} =
...
   result.setIfNeeded("INJECTOR_COMMIT_ID", getChalkCommitId())
   result.setIfNeeded("INJECTOR_ENV", getEnvDict())
-  result.setIfNeeded("DATE_CHALKED", pack(getDate()))
-  result.setIfNeeded("TIME_CHALKED", pack(getTime()))
-  result.setIfNeeded("TZ_OFFSET_WHEN_CHALKED", pack(getOffset()))
-  result.setIfNeeded("DATETIME_WHEN_CHALKED", pack(getDateTime()))
+  result.setIfNeeded("DATE_CHALKED", startTime.utc.format(timesDateFormat))
+  result.setIfNeeded("TIME_CHALKED", startTime.utc.format(timesTimeFormat))
+  result.setIfNeeded("TZ_OFFSET_WHEN_CHALKED", startTime.utc.format(timesTzFormat))
+  result.setIfNeeded("DATETIME_WHEN_CHALKED", startTime.utc.format(timesIso8601Format))
```
mbaltrusitis, line 262: 👍

mbaltrusitis (dismissed):

tests/functional/utils/git.py
```diff
@@ -15,7 +15,9 @@
 
 logger = get_logger()
 
-DATE_FORMAT = re.compile(r"^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3}([+-]\d{4}|Z)$")
+DATE_FORMAT = re.compile(
+    r"^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3}([+-]\d{2}:\d{2}|Z)$"
```
mbaltrusitis, line 19: e/ cracks RegEx knuckles

```suggestion
"^(?P<year>\d{4})-(?P<month>0[1-9]|1[0-2])-(?P<day>0[1-9]|[12]\d|3[01])T(?P<hour>[01]\d|2[0-3]):(?P<minute>[0-5]\d):(?P<second>[0-5]\d)(?:\.(?P<fraction>\d+))?(?P<timezone>Z|(?P<offset_sign>[+-])(?P<offset_hour>[01]\d|2[0-3]):(?P<offset_minute>[0-5]\d))$"
```

- tightened up regex to disallow impossible dates and times
- named capture groups to self-document (and aide?) in comprehension at the cost of performance. if this pattern is ever used in the actual app, non-capture groups are suggested

image

mbaltrusitis (dismissed):

tests/functional/utils/git.py
miki725: python actually supports documenting regex in-line:

https://docs.python.org/3.13/library/re.html#re.X

```python
a = re.compile(r"""\d +  # the integral part
                   \.    # the decimal point
                   \d *  # some fractional digits""", re.VERBOSE)
```

but I just pivoted to use stdlib `fromisoformat()` method vs attempting to validate dates via regex..

eventually if we need to will also make it easy to add assertions about ranges of dates/etc but for now it simply validates something is valid Iso8601 string...

[`f5063f9` (#458)](https://github.com/crashappsec/chalk/pull/458/commits/f5063f91dbab64fd78bd988c8b47a504462e33aa)

PR #459, opened 2024-12-12
fix: context manager to scope logs collection for _OP_ERRORS and ERR_INFO

miki725:

## Issue

not all logs were showing up in `_OP_ERRORS`

## Description

without context manager it was a manual process which was setting the current error object which is responsible for populating either _OP_ERRORS or ERR_INFO keys. As such for example during docker build, as soon as chalk was starting subscan of the docker binary, its target was switched to the docker binary chalk object hence no more logs were collected for `_OP_ERRORS` even after subscan was complete.

Now operations explicitly have to use context manager to limit the scope of the current error object hence guaranteeing scope is always reverted to the original object. Docker collection is more involved due to multi-platform builds as a single base-chalk is used to create possibly many subplatform chalk objects hence the context manager is necessary in a few surgical places.

## Testing

```
➜ reset; CHALK_BUILD=debug make; and echo ------------------------------; and ./chalk docker build -f (echo FROM alpines | psub) .
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -190,6 +190,18 @@
   wrong commit TZ was being reported as committed in git which was not correct.
   ([#458](https://github.com/crashappsec/chalk/pull/458))
 
+- `_OP_ERRORS` includes all logs from chalkmark `ERR_INFO`,
+  even when its collection fails
+  ([#459](https://github.com/crashappsec/chalk/pull/459))
+
+- `docker buildx build` without both `--push` or `--load` report their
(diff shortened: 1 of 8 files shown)


miki725:
https://app.shortcut.com/crashoverride-1/story/3299/chalk-fallback-does-not-include-logs-from-failed-chalkmark

mbaltrusitis (dismissed):

mbaltrusitis (approved):

PR #461, opened 2024-12-24
feat: adding _DOCKER_USED_REGISTRIES key

miki725:

## Issue

report metadata about used registries during docker operations

## Description

This is for the most part just exposing the info already stored in the registry cache.

At some point we might want to collect all registries, even the ones not used throughout the docker command but for now this is a good start.

Also reporting builder info/configs via:

* `_DOCKER_BUILDER_NODES_INFO`
* `_DOCKER_BUILDER_NODES_CONFIG`

## Testing

trigger any build within tests image which already has a bunch of buildx configs:

```
➜ reset; make; and IP=10.10.20.30 docker compose run -it --rm -v /etc/chalk.c4m:/etc/chalk.c4m tests $PWD/chalk docker buildx build -f (echo -e 'FROM ghcr.io/sigstore/cosign/cosign:v2.2.3 AS cosign\nFROM alpine\nCOPY --from=cosign /ko-app/cosign /cosign' | psub) . --push -t 10.10.20.30:5045/secure -t 10.10.20.30:5046/insecure -t 10.10.20.30:5044/http
```


miki725:
https://app.shortcut.com/crashoverride-1/story/2336/chalk-report-registry-config

viega (approved):

mbaltrusitis (dismissed):
LGTM :christmas_tree:

src/run_management.nim
```diff
@@ -198,53 +198,49 @@ proc newChalk*(name:         string            = "",
...
+  when T is Box:
+    setIfNotEmptyBox(o, k, v)
+  elif T is JsonNode:
+    if v != nil:
+      setIfNotEmptyBox(o, k, v.nimJsonToBox())
+  elif T is Option:
+    if v.isSome():
+      setIfNotEmptyBox(o, k, pack(v.get()))
+  else:
+    setIfNotEmptyBox(o, k, pack(v))
```
mbaltrusitis, line 219: is this better as compound if statements to eliminate the nested conditionals?

tests/functional/test_plugins.py
```diff
@@ -21,6 +21,7 @@
 logger = get_logger()
 
 
+@pytest.mark.flaky(max_runs=5)
```
mbaltrusitis, line 24: do we know what specifically makes this flaky? would make a good note/comment

src/run_management.nim
miki725: the outer is compile-time conditional. note the `when`vs `if`

https://nim-docs.readthedocs.io/en/latest/manual/statements_expressions/#when-statement

note that `when` operates on a generic type `T` here which is not available at run-time as its a compile-time construct

for example another common use-case for `when` is for debug-related things:

https://github.com/crashappsec/chalk/blob/81c0d04646a1ed9aee623f13edf469e661d3b2a9/src/chalk_common.nim#L507-L517

when doing release builds, that code-block is completely omitted from the binary

tests/functional/test_plugins.py
miki725: unfortunately no. I think there is a bug somewhere in chalk as the timeout is very generious now however so far Im unable to replicate it locally but it does happen with some frequency in CI which is rather annoying

mbaltrusitis (approved):

PR #462, opened 2025-01-03
fix: empty buildx node config

miki725:

## Issue

possible sefault with empty builder node config

## Description

not all build nodes have a config file associated with them and as such buildkitd.toml might not exist. That doesnt mean the node doesnt exist, it simply means it has no custom config.

For example:

```
docker buildx create test
```

To correctly detect whether a node has a config or not, the container entrypoint/cmd are inspected to find the corresponding `--config` and if present, only then attempts to read the node config. Otherwise empty config is used.

This also surfaced a segfault when using empty config the code was assuming ["registry"] was always present which is not the case.

As the node code now needed to inspect the container args, moved node-related logic to separate nodes.nim therefore allowing it to reference inspect.nim as otherwise would result in circular import as inspect.nim needs exe.nim to make docker subprocess calls.

## Testing

```
➜ maketest test_docker.py::test_build[cwd0-None-False-True-True-empty_builder] -x --pdb
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -256,7 +256,7 @@
   {
     "_DOCKER_USED_REGISTIES" {
       "example.com:5044": {
-        "url": "https://example.com:5044/v2",
+        "url": "https://example.com:5044/v2/",
         "source": "buildx",
         "scheme": "https",
         "http": false,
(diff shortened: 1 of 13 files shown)


mbaltrusitis (approved):
LGTM

PR #476, opened 2025-01-28
refactor: enforce consistent usages across codebase

miki725:
now however a symbol is defined, it must be used with the case case so no more `fooBar` vs `foobar` vs `foo_bar` typos which make grepping much more complicated


viega (approved):
Did you automate looking for such things?  This is probably my least favorite part of Nim :)

miki725:
yep

https://github.com/crashappsec/chalk/blob/58fa4d94a52141968b3d0a55a273bcf2192485e1/config.nims#L24-L30

from now on any incosistent usage (within chalk itself) will fail compilation 🎉

PR #486, opened 2025-02-11
fix: check buildkit frontend version to determine context arg compatibility

miki725:

## Issue

Builds which specify frontend can sometimes fallback due to incompatible `--build-arg`

## Description

Not all older frontends support `--build-arg` and sometimes they do no specify the version as part of the syntax directive. For example:

```
# syntax: docker/dockerfile:experimental
```

Chalk cannot infer anything about the meaning of `experimental` as its not pinned to any version. Therefore we call the dockerfile frontend container and check its actual version.
This is possibly not as efficient even if the version is specified however that would also require some heuristics. For example:

```
# syntax: docker/dockerfile:1
```

As `1` can change over time it might not be clear which syntax should actually be honored. Therefore it is probably a lot simpler to simply ask it what version it is instead of possibly error-prone version detection heuristics.

## Testing

```
➜ maketest test_docker.py::test_build -x
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,13 @@
   on `hello.py`. Now chalk will compute nearest `git` repository
   and run external tools on it instead.
   ([#485](https://github.com/crashappsec/chalk/pull/485))
+- When `Dockerfile` specifies syntax directive, chalk checks buildkit
+  frontend version compatibility as older frontends do not support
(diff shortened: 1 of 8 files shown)


miki725:
https://app.shortcut.com/crashoverride-1/story/4161/chalk-docker-build-fails-causing-fallback-to-vanilla-build-which-succeeds

mbaltrusitis (approved):

PR #490, opened 2025-02-19
ci: remove connect.c4m [ci skip]

miki725:
this was deprecated for a long time already after chalk profiles shipped


MyNameIsMeerkat (approved):
Great to have this cleaned up 🙏

PR #498, opened 2025-03-18
feat: chalk insert lambda zip mode

mbaltrusitis:
- [ ] Followed the steps in the contributor's guide: https://crashoverride.com/docs/other/contributing#filing-the-pull-request
- [ ] PR title uses [semantic commit messages](https://nitayneeman.com/posts/understanding-semantic-commit-messages-using-git-and-angular/#fix)
- [ ] Filled out the template to a useful degree
- [ ] Updated `CHANGELOG.md` if necessary

## Issue

https://app.shortcut.com/crashoverride-1/epic/4279?group_by=none&ct_workflow=all&cf_workflow=500000306

## Description

Initial approach for injecting `chalk` binary inside of a lambda zip archive. After a round of feedback I think it'll make sense to refactor some of the zip archive operations to `codecZip.nim`.

### to do

- [ ] get initial feedback
- [ ] fix executable bit issue
- [ ] (potentially) refactor into `codecZip.nim`

## Testing

`make tests args="test_lambda.py"`

### DIFF
--- a/src/configs/chalk.c42spec
+++ b/src/configs/chalk.c42spec
@@ -2920,6 +2920,16 @@ Currently, the only available tool out of the box is trufflehog.
 """
   }
 
+  field lambda_mode {
+    type:     bool
+    default:  false
+    shortdoc: "lambda mode"
+    doc:      """
+When present, inject the executed chalk binary itself in the zip
+archive being processed upon insertion.
+"""
+  }
+
   field recursive {
     type:     bool
     default:  true
--- a/tests/functional/chalk/runner.py
+++ b/tests/functional/chalk/runner.py
@@ -433,6 +433,7 @@ class Chalk:
         expecting_report: bool = True,
         expecting_chalkmarks: bool = True,
         use_embedded: bool = True,
+        params: Optional[list[str]] = None,
     ) -> ChalkProgram:
         result = self.run(
             command="insert",
@@ -444,6 +445,7 @@ class Chalk:
             ignore_errors=ignore_errors,
             expecting_report=expecting_report,
             use_embedded=use_embedded,
+            params=params,
         )
         if expecting_report:
             if expecting_chalkmarks:
--- a/tests/functional/test_lambda.py
+++ b/tests/functional/test_lambda.py
@@ -1,0 +1,186 @@
+# Copyright (c) 2025, Crash Override, Inc.
+#
+# This file is part of Chalk
+# (see https://crashoverride.com/docs/chalk)
+from pathlib import Path
+import shutil
+import zipfile
+
+import pytest
+
+from .chalk.runner import Chalk
+from .conf import ZIPS
+from .utils.log import get_logger
+
+
+logger = get_logger()
+
+
+# Test data fixtures
+PYTHON_LAMBDA_ZIP = ZIPS / "python" / "my_deployment_package.zip"
+NODEJS_LAMBDA_ZIP = ZIPS / "nodejs" / "function.zip"
+GOLANG_LAMBDA_ZIP = ZIPS / "golang" / "myFunction.zip"
+MISC_LAMBDA_ZIP = ZIPS / "misc" / "misc.zip"
+EMPTY_ZIP = ZIPS / "empty" / "empty.zip"
+
+
+def verify_chalk_binary_in_zip(zip_path: Path, extract_dir: Path):
+    """Helper function to verify chalk binary exists in the zip file."""
+    # Extract the zip contents
+    extract_dir.mkdir(exist_ok=True)
+
+    with zipfile.ZipFile(zip_path, "r") as zip_ref:
+        zip_ref.extractall(extract_dir)
+
+    # Verify the chalk binary was added
+    chalk_binary_path = extract_dir / "chalk"
+    assert chalk_binary_path.exists(), "Chalk binary was not found in the zip"
+    # TODO: uncomment when zip archive exec perms if fixed
+    # assert os.access(chalk_binary_path, os.X_OK), "Chalk binary is not executable"
+
+
+def verify_chalk_mark_added(result, test_file):
+    """Helper function to verify chalk mark was added to the file."""
+    assert result.report.marks_by_path.contains(
+        {str(test_file): {}}
+    ), "Chalk mark was not added to the zip file"
+
+
+@pytest.mark.parametrize(
+    "copy_files",
+    [
+        [PYTHON_LAMBDA_ZIP],
+        [NODEJS_LAMBDA_ZIP],
+    ],
+    indirect=True,
+)
+def test_lambda_insert(
+    tmp_data_dir: Path,
+    chalk: Chalk,
+    copy_files: list[Path],
+):
+    """Test `chalk insert --lambda` adds the binary to the zip."""
+    test_file = copy_files[0]
+
+    # Run chalk insert with the --lambda
+    insert = chalk.insert(
+        artifact=tmp_data_dir,
+        virtual=False,
+        params=["--lambda"],
+    )
+
+    # Verify the chalk mark was added
+    verify_chalk_mark_added(insert, test_file)
+
+    # Verify chalk binary was added to the zip
+    extract_dir = tmp_data_dir / "extracted"
+    verify_chalk_binary_in_zip(test_file, extract_dir)
+
+    # Cleanup
+    shutil.rmtree(extract_dir)
+
+
+def test_lambda_insert_empty_zip(
+    tmp_data_dir: Path,
+    chalk: Chalk,
+):
+    """Test `chalk insert --lambda` on an empty zip file.
+
+    Empty zip files should be handled gracefully - the command should
+    complete with exit code 0 but no chalk marks should be created.
+    """
+
+    # Run chalk insert with the --lambda
+    insert = chalk.insert(
+        artifact=tmp_data_dir,
+        virtual=False,
+        params=["--lambda"],
+        expecting_chalkmarks=False,  # Don't expect marks for empty zip
+    )
+
+    # Verify command succeeded but no marks were created
+    assert insert.exit_code == 0, "Command should have exited with code 0"
+    assert (
+        "_CHALKS" not in insert.report
+    ), "No chalks should be created for empty zip file"
+    assert (
+        "_OP_CHALK_COUNT" in insert.report and insert.report["_OP_CHALK_COUNT"] == 0
+    ), "Chalk count should be 0"
+
+
+@pytest.mark.parametrize(
+    "copy_files",
+    [
+        [PYTHON_LAMBDA_ZIP],
+    ],
+    indirect=True,
+)
+def test_lambda_extract_after_insert(
+    tmp_data_dir: Path,
+    chalk: Chalk,
+    copy_files: list[Path],
+):
+    """Test extraction works correctly after insertion.
+
+    After inserting chalk with --lambda flag, we should be able to
+    extract the chalk marks without errors, and the marks should
+    identify the file as a ZIP artifact.
+    """
+    test_file = copy_files[0]
+
+    # First insert a chalk mark with lambda flag
+    insert = chalk.insert(
+        artifact=tmp_data_dir,
+        virtual=False,
+        params=["--lambda"],
+    )
+    verify_chalk_mark_added(insert, test_file)
+
+    # Now extract and verify we can read the chalk mark
+    extract = chalk.extract(artifact=tmp_data_dir, ignore_errors=True)
+    verify_chalk_mark_added(extract, test_file)
+
+    # chalk_binary_path = tmp_data_dir / "chalk"
+    # Verify ZIP artifact type is detected
+    for mark in extract.marks:
+        assert mark.has(ARTIFACT_TYPE="ZIP"), "ZIP artifact type not detected"
+    # TODO: uncomment when zip archive exec perms if fixed
+    # assert os.access(chalk_binary_path, os.X_OK), "Chalk binary is not executable"
+
+
+@pytest.mark.parametrize(
+    "copy_files",
+    [
+        [GOLANG_LAMBDA_ZIP],
+        [MISC_LAMBDA_ZIP],
+    ],
+    indirect=True,
+)
+def test_lambda_with_other_zip_formats(
+    tmp_data_dir: Path,
+    chalk: Chalk,
+    copy_files: list[Path],
+):
+    """Test the --lambda flag with different zip file formats.
+
+    The lambda insertion should work with various zip formats commonly
+    used for different lambda runtimes (Go, misc).
+    """
+    test_file = copy_files[0]
+
+    # Run chalk insert with the lambda flag
+    insert = chalk.insert(
+        artifact=tmp_data_dir,
+        virtual=False,
+        params=["--lambda"],
+    )
+
+    # Verify the chalk mark was added
+    verify_chalk_mark_added(insert, test_file)
+
+    # Verify chalk binary was added to the zip
+    extract_dir = tmp_data_dir / "extracted"
+    verify_chalk_binary_in_zip(test_file, extract_dir)
+
+    # Cleanup
+    shutil.rmtree(extract_dir)
--- a/README.md
+++ b/README.md
@@ -4,9 +4,13 @@
 
 ## About Chalk
 
-Chalk™ captures metadata at build time, and can add a small 'chalk mark' (metadata) to any artifacts, so they can be identified in production. Chalk can also extract chalk marks and collect additional metadata about the operating environment when it does this.
+Chalk™ captures metadata at build time, and can add a small 'chalk mark' (metadata) to any
+artifacts, so they can be identified in production. Chalk can also extract chalk marks and collect
+additional metadata about the operating environment when it does this.
 
-Using Chalk, you can build a graph connecting people, development, builds and production, so that devops engineers understand what is happening in the development process, and so that developers can understand what is happening in the infrastructure.
+Using Chalk, you can build a graph connecting people, development, builds and production, so that
+devops engineers understand what is happening in the development process, and so that developers
+can understand what is happening in the infrastructure.
 
 ## How-tos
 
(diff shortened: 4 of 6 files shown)


tests/functional/test_lambda.py
```diff
@@ -0,0 +1,186 @@
...
+    """Helper function to verify chalk binary exists in the zip file."""
+    # Extract the zip contents
+    extract_dir.mkdir(exist_ok=True)
+
+    with zipfile.ZipFile(zip_path, "r") as zip_ref:
+        zip_ref.extractall(extract_dir)
+
+    # Verify the chalk binary was added
+    chalk_binary_path = extract_dir / "chalk"
+    assert chalk_binary_path.exists(), "Chalk binary was not found in the zip"
```
miki725, line 37: in pytest adding explicit assertion message is not necessary as it prints useful context about the error:

given:

```python
def a():
    return 1
def b():
    return 2
def test_foo():
    assert a() == b()
```

error will be:

```python
    def test_foo():
>       assert a() == b()
E       assert 1 == 2
E        +  where 1 = a()
E        +  and   2 = b()
```

tests/functional/test_lambda.py
```diff
@@ -0,0 +1,186 @@
...
+    # TODO: uncomment when zip archive exec perms if fixed
+    # assert os.access(chalk_binary_path, os.X_OK), "Chalk binary is not executable"
+
+
+def verify_chalk_mark_added(result, test_file):
```
miki725, line 42: no type annotations which we try to use everywhere in chalk tests

tests/functional/test_lambda.py
```diff
@@ -0,0 +1,186 @@
...
+
+    # Verify the chalk mark was added
+    verify_chalk_mark_added(insert, test_file)
+
+    # Verify chalk binary was added to the zip
+    extract_dir = tmp_data_dir / "extracted"
+    verify_chalk_binary_in_zip(test_file, extract_dir)
+
+    # Cleanup
+    shutil.rmtree(extract_dir)
```
miki725, line 80:
```suggestion
```

this is all already being cleaned up via the `tmp_data_dir` fixture:

https://github.com/crashappsec/chalk/blob/d332e41174a714503b84434556c2970e2b980a62/tests/functional/conftest.py#L84-L88

tests/functional/test_lambda.py
```diff
@@ -0,0 +1,186 @@
...
+    # Run chalk insert with the --lambda
+    insert = chalk.insert(
+        artifact=tmp_data_dir,
+        virtual=False,
+        params=["--lambda"],
+        expecting_chalkmarks=False,  # Don't expect marks for empty zip
+    )
+
+    # Verify command succeeded but no marks were created
+    assert insert.exit_code == 0, "Command should have exited with code 0"
```
miki725, line 102:
```suggestion
```

this is already checked by `chalk.insert()` and `chalk.run()`:

https://github.com/crashappsec/chalk/blob/d332e41174a714503b84434556c2970e2b980a62/tests/functional/chalk/runner.py#L389-L392

in general the runner wrapper does a whole bunch of sanity checks whenever it calls chalk

tests/functional/test_lambda.py
```diff
@@ -0,0 +1,186 @@
...
+    Empty zip files should be handled gracefully - the command should
+    complete with exit code 0 but no chalk marks should be created.
+    """
+
+    # Run chalk insert with the --lambda
+    insert = chalk.insert(
+        artifact=tmp_data_dir,
+        virtual=False,
+        params=["--lambda"],
+        expecting_chalkmarks=False,  # Don't expect marks for empty zip
```
miki725, line 98: this chalkmark checks what is present in the chalk report. its not related to the content in the zip file. this should not be disabled unless the report is expected not to have any chalkmarks in the report such as running chalk on empty folder  - `chalk insert <empty-folder>`

tests/functional/test_lambda.py
```diff
@@ -0,0 +1,186 @@
...
+):
+    """Test `chalk insert --lambda` on an empty zip file.
+
+    Empty zip files should be handled gracefully - the command should
+    complete with exit code 0 but no chalk marks should be created.
+    """
+
+    # Run chalk insert with the --lambda
+    insert = chalk.insert(
+        artifact=tmp_data_dir,
```
miki725, line 95: 😕 description does not matchwhat the test case does. it runs on empty empty folder which is not related to empty zip files

tests/functional/test_lambda.py
```diff
@@ -0,0 +1,186 @@
...
+    )
+
+    # Verify command succeeded but no marks were created
+    assert insert.exit_code == 0, "Command should have exited with code 0"
+    assert (
+        "_CHALKS" not in insert.report
+    ), "No chalks should be created for empty zip file"
+    assert (
+        "_OP_CHALK_COUNT" in insert.report and insert.report["_OP_CHALK_COUNT"] == 0
+    ), "Chalk count should be 0"
```
miki725, line 108: these are already handled by these options:

https://github.com/crashappsec/chalk/blob/d332e41174a714503b84434556c2970e2b980a62/tests/functional/chalk/runner.py#L433-L434

they will assert appropriate values in the report are present

tests/functional/test_lambda.py
```diff
@@ -0,0 +1,186 @@
...
+    # First insert a chalk mark with lambda flag
+    insert = chalk.insert(
+        artifact=tmp_data_dir,
+        virtual=False,
+        params=["--lambda"],
+    )
+    verify_chalk_mark_added(insert, test_file)
+
+    # Now extract and verify we can read the chalk mark
+    extract = chalk.extract(artifact=tmp_data_dir, ignore_errors=True)
```
miki725, line 140: why ignore errors? this should only be used for commands where we legitimately expect errors such as trying to load bad chalk config in tests

tests/functional/test_lambda.py
```diff
@@ -0,0 +1,186 @@
...
+    verify_chalk_mark_added(extract, test_file)
+
+    # chalk_binary_path = tmp_data_dir / "chalk"
+    # Verify ZIP artifact type is detected
+    for mark in extract.marks:
+        assert mark.has(ARTIFACT_TYPE="ZIP"), "ZIP artifact type not detected"
```
miki725, line 146: this is already tested by `extract` util

https://github.com/crashappsec/chalk/blob/d332e41174a714503b84434556c2970e2b980a62/tests/functional/chalk/runner.py#L494-L500

tests/functional/test_lambda.py
```diff
@@ -0,0 +1,186 @@
...
+
+@pytest.mark.parametrize(
+    "copy_files",
+    [
+        [GOLANG_LAMBDA_ZIP],
+        [MISC_LAMBDA_ZIP],
+    ],
+    indirect=True,
+)
+def test_lambda_with_other_zip_formats(
```
miki725, line 159: not following intent here?

if we want chalk to be able to support multiple zip formats, we should just have a single test case with parametrize inputs which will run the test case on all the variants. from the test case perspective it should not matter how zip was created

miki725 (changes requested):

tests/functional/chalk/runner.py
```diff
@@ -433,6 +433,7 @@ def insert(
         expecting_report: bool = True,
         expecting_chalkmarks: bool = True,
         use_embedded: bool = True,
+        params: Optional[list[str]] = None,
```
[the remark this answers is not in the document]
mbaltrusitis: totally makes sense. thanks for explaining the convention.

tests/functional/test_lambda.py
```diff
@@ -0,0 +1,145 @@
...
+
+# Test data fixtures
+PYTHON_LAMBDA_ZIP = ZIPS / "python" / "my_deployment_package.zip"
+NODEJS_LAMBDA_ZIP = ZIPS / "nodejs" / "function.zip"
+GOLANG_LAMBDA_ZIP = ZIPS / "golang" / "myFunction.zip"
+MISC_LAMBDA_ZIP = ZIPS / "misc" / "misc.zip"
+EMPTY_ZIP = ZIPS / "empty" / "empty.zip"
+
+
+def verify_chalk_binary_in_zip(zip_path: Path, extract_dir: Path):
```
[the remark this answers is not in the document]
mbaltrusitis: this was left over from when i was debugging. removed in latest update.

tests/functional/test_lambda.py
```diff
@@ -0,0 +1,145 @@
...
+
+    # First insert a chalk mark with --inject-binary flag
+    insert = chalk.insert(artifact=tmp_data_dir, virtual=False, inject_binary=True)
+    assert insert.marks_by_path.contains({str(test_file): {}})
+    assert insert.mark.contains({"_OP_ARTIFACT_TYPE": "ZIP"})
+
+    # Now extract and verify we can read the chalk mark
+    extract = chalk.extract(artifact=tmp_data_dir)
+    assert extract.marks_by_path.contains({str(test_file): {}})
+    assert extract.mark.contains({"_OP_ARTIFACT_TYPE": "ZIP"})
```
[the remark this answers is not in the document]
mbaltrusitis: Since the set of keys is not consistent across insert and extract, I've updated the test to assert equality across their common keys.

src/plugins/codecZip.nim
```diff
@@ -177,6 +185,56 @@ proc zipScan*(self: Plugin, loc: string): Option[ChalkObj] {.cdecl.} =
...
+  try:
+    # Get the currently executing chalk binary path
+    let myAppPath = getMyAppPath()
+
+    let chalkBinaryContent = tryToLoadFile(myAppPath)
+
+    var zipCache: ZipCache
+    try:
+      zipCache = cast[ZipCache](chalk.cache)
+      if zipCache == nil or zipCache.tmpDir == "":
```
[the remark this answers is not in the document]
mbaltrusitis: removed :+1:

src/plugins/codecZip.nim
```diff
@@ -177,6 +185,56 @@ proc zipScan*(self: Plugin, loc: string): Option[ChalkObj] {.cdecl.} =
 
     return some(chalk)
 
+proc insertChalkBinaryIntoZip*(chalk: ChalkObj): bool =
+  ## Inserts the chalk binary into a zip archive
+  ## Returns true if successful, false otherwise
+
+  result = false
+
+  if chalk.myCodec == nil or chalk.myCodec.name != "zip" or chalk.cache == nil:
```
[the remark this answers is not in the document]
mbaltrusitis: removed :+1:

src/plugins/codecZip.nim
```diff
@@ -205,6 +263,13 @@ proc doZipWrite(chalk: ChalkObj, encoded: Option[string], virtual: bool) =
 
 proc zipHandleWrite*(self: Plugin, chalk: ChalkObj, encoded: Option[string])
                    {.cdecl.} =
+  let injectBinary = attrGet[bool]("inject_binary")
+  if injectBinary:
+    info(chalk.name & ": inserting binary into zip archive")
+
+    if not insertChalkBinaryIntoZip(chalk):
+      error(chalk.name & ": failed to insert chalk binary into zip archive")
```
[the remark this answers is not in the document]
mbaltrusitis: removed the try/except blocks and handled it on the `doZipWrite` func

src/configs/chalk.c42spec
```diff
@@ -3145,6 +3145,16 @@ Currently, the only available tool out of the box is trufflehog.
 """
   }
 
+  field inject_binary {
```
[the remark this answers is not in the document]
miki725: thoughts? ^^

src/plugins/codecZip.nim
```diff
@@ -177,6 +185,41 @@ proc zipScan*(self: Plugin, loc: string): Option[ChalkObj] {.cdecl.} =
...
+    if contentSuccess:
+      # Make executable
+      contentTargetPath.makeExecutable()
+      trace(chalk.name & ": added chalk binary to content zip dir")
+
+      result = true
+    else:
+      error(chalk.name & ": failed to add chalk binary to zip directory")
+  else:
+    error(chalk.name & ": zip content directory does not exist")
```
miki725, line 219:
```suggestion
    raise newException(ValueError, "zip content directory does not exist")
```

tests/functional/test_zip.py
```diff
@@ -96,3 +104,111 @@ def test_valid(
...
+
+@pytest.mark.parametrize(
+    "copy_files",
+    [
+        [PYTHON_LAMBDA_ZIP],
+        [NODEJS_LAMBDA_ZIP],
+    ],
+    indirect=True,
+)
+def test_lambda_insert(
```
miki725, line 124: these basically duplicate already existing zip tests:

*  only addition is it should be testing is the binary injection
* currently these `lambda_*` tests dont follow same semantics as other tests. for example some of these zips are slower to process hence those tests are marked as `@slow()` which is missing here
 
both are easy to achieve via pytest params. something like this should do the trick:

```diff
diff --git tests/functional/test_zip.py tests/functional/test_zip.py
index 0cd9e7b..48bab7c 100644
--- tests/functional/test_zip.py
+++ tests/functional/test_zip.py
@@ -23,19 +23,26 @@ logger = get_logger()
     ],
     indirect=True,
 )
+@pytest.mark.parametrize("inject_binary", [True, False])
 @pytest.mark.parametrize("virtual", [True, False])
 def test_valid_slow(
     tmp_data_dir: Path,
     chalk: Chalk,
     copy_files: list[Path],
     virtual: bool,
+    inject_binary: bool,
 ):
     test_file = copy_files[0]

     # chalk reports generated by insertion, json array that has one element
-    insert = chalk.insert(artifact=tmp_data_dir, virtual=virtual)
+    insert = chalk.insert(
+        artifact=tmp_data_dir, virtual=virtual, inject_binary=inject_binary
+    )
     assert insert.report.marks_by_path.contains({str(test_file): {}})

+    if inject_binary:
+        verify_chalk_binary_in_zip(test_file)
+
     # array of json chalk objects as output, of which we are only expecting one
     extract = chalk.extract(artifact=tmp_data_dir, virtual=virtual)
     if not virtual:
```

you can do equivalent diff for all the tests here

tests/functional/test_lambda.py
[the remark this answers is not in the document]
miki725: as these tests were moved to `test_zip.py`, this test file should be removed

src/plugins/codecZip.nim
```diff
@@ -177,6 +185,41 @@ proc zipScan*(self: Plugin, loc: string): Option[ChalkObj] {.cdecl.} =
...
+
+  return result
```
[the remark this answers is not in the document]
mbaltrusitis: i've been including it to have an explicit notation of what is returned. I'll remove to keep it more idiomatic and consistent.

src/plugins/codecZip.nim
mbaltrusitis: Refactored both `if`s to raise if not true.

tests/functional/test_zip.py
```diff
@@ -73,26 +90,85 @@ def test_empty(
...
+
+    # Now extract and verify we can read the chalk mark
+    extract = chalk.extract(artifact=tmp_data_dir)
+    assert extract.marks_by_path.contains({str(test_file): {}})
+    assert extract.mark.contains({"_OP_ARTIFACT_TYPE": "ZIP"})
+    # keys common between `insert` and `extract` are equal
+    assert all(
+        insert.mark[key] == extract.mark[key]
+        for key in insert.mark.keys() & extract.mark.keys()
+    )
```
miki725, line 159: I thought we will just add the `inject_binary` param to all other tests with the additional verify_chalk_binary_in_zip` check

```suggestion
```

most of the assertions here are already covered by above tests:

* `_OP_ARTIFACT_TYPE`  is asserted as part of `chalk.extract` by default:

https://github.com/crashappsec/chalk/blob/84f1ac4a9f7eb51102653127dc21ba9fac4b4452/tests/functional/chalk/runner.py#L508-L510

the only additional assertion here which is missing in the above tests is comparing `insert.mark` and `extract.mark`. we can probably just do that assertion in above tests. although that can be done much simpler:

```python
assert insert.mark.contains(extract.mark)
```

that will show pretty errors too whenever any subkey wont match... see:

https://github.com/crashappsec/chalk/blob/84f1ac4a9f7eb51102653127dc21ba9fac4b4452/tests/functional/utils/dict.py#L178-L181

no need to do another pass of `chalk.insert` just for one additional assertion. keep in mind that these are functional and have quite a bit of overhead. each test case pytest has to:

* setup tmp folder
* copy vanilla chalk into tmp folder
* load into it testing config

and then each chalk operation makes a subprocess call which is not very fast in general so these things add up. test suite is already 10+ minutes. over time went from like 3 minutes. at some point we might need to optimize this a bunch of mark some tests `@slow`....

src/plugins/codecZip.nim
```diff
@@ -177,6 +185,28 @@ proc zipScan*(self: Plugin, loc: string): Option[ChalkObj] {.cdecl.} =
...
+  # Check the directory exists
+  if not dirExists(contentDir):
+      raise newException(IOError, chalk.name & ": zip content directory does not exist")
+
+  # Write the chalk binary to content directory
+  if not tryToWriteFile(contentTargetPath, chalkBinaryContent):
+    raise newException(ValueError, "failed to add chalk binary to zip directory")
+
+    contentTargetPath.makeExecutable()
+    trace(chalk.name & ": added chalk binary to content zip dir")
```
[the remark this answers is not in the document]
miki725: 🛑

miki725 (commented):
from my side its in good shape. some todos:

- cleanup bad indent in the zip codec. currently one section is unreachable. see comment in there
- update changelog with PR link as per our convention
- unnecessary test cleanup

CHANGELOG.md
```diff
@@ -70,6 +70,10 @@
   ([#515](https://github.com/crashappsec/chalk/pull/515))
 - Reporting all IMDSv2 errors in `FAILED_KEYS` or `_OP_FAILED_KEYS`.
   ([#519](https://github.com/crashappsec/chalk/pull/519))
+- Adds `--inject-binary` to the `chalk insert` command
```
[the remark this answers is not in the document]
miki725: now change log change seems to be missing? accidental delete?

tests/functional/test_zip.py
```diff
@@ -73,26 +94,45 @@ def test_empty(
...
+        assert extract.mark.contains(insert.mark.if_exists())
+    if inject_binary_into_zip:
+        assert does_contain_chalk_binary_in_zip(test_file, virtual)
+
+
+def does_contain_chalk_binary_in_zip(zip_path: Path, virtual: bool):
+    """Helper function to verify chalk binary exists in the zip file."""
+    if virtual:
+        logger.info("virtual, return true")
+        return True
```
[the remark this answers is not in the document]
mbaltrusitis: fixed and renamed the utility

miki725:
@mbaltrusitis before merging, please update the PR description with all the template bits

PR #500, opened 2025-03-25
feat: object store per key in reports

miki725:

## Issue

Chalk reports can be very large which send duplicate data across builds (e.g. sbom).

## Description

Instead of reporting raw value in the report template, each key could be uploaded to an object store and its reference be included in the report. This allows to deduplicate large keys independently to object store and then refer the same object across reports.

To use it object store needs to be configured:

```
object_store_config foo {
  object_store: "presign"
  object_store_presign {
    uri: "http://localhost:8080/v0.1/objects"
  }
}
```

Then per report template the object store can be used for specific keys:

```
report_template foo {
  key.SBOM.object_store = "foo"
}
```

Currently only `presign` object store is implemented however the infrastructure is ready to add more object stores in the future such as `s3`, `file`, etc.

## Testing

```
➜ maketest test_plugins.py::test_syft_docker --pdb --logs
```

## TODO

- [x] doc updates
- [x] tests
- [x] handle race conditions

### DIFF
--- a/src/object_store/api.nim
+++ b/src/object_store/api.nim
@@ -6,7 +6,7 @@
 ##
 
 import std/[tables, uri]
-import ".."/[config]
+import ".."/[config, normalize]
 import "."/[presign]
 
 let objectStores = {
@@ -34,11 +34,15 @@ proc getObjectStoreConfigByName(name: string): ObjectStoreConfig =
   result = store.init(store, name)
   objectStoreConfigs[name] = result
 
-proc newObjectStoreRef*(self: ObjectStoreConfig, k: string, data: string): ObjectStoreRef =
+proc newObjectStoreRef*(self:          ObjectStoreConfig,
+                        k:             string,
+                        data:          string,
+                        canonicalData: Box,
+                        ): ObjectStoreRef =
   return ObjectStoreRef(
     config: self,
     key:    k,
-    id:     data.sha256Hex(),
+    id:     canonicalData.binEncodeItem().sha256Hex(),
     digest: data.sha256Hex(),
   )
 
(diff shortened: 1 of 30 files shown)


miki725:
https://app.shortcut.com/crashoverride-1/story/5737/implement-object-store-config-in-chalk

src/object_store/api.nim
```diff
@@ -0,0 +1,148 @@
...
+proc canonicalizeKey(key: string, data: Box): Box =
+  let callbackOpt = attrGetOpt[CallbackObj]("keyspec." & key & ".canonicalize")
+  if callbackOpt.isNone():
+    trace("object store: no canonicalize() for " & key)
+    return data
+  let callback = callbackOpt.get()
+  trace("object store: canonicalizing " & key & " with " & $callback)
+  try:
+    let canonicalized = runCallback(callback, @[data])
+    #let canonicalized = runCallback("keyspec." & key & ".canonicalize", @[data])
```
mbaltrusitis, line 69: remove?

src/object_store/presign.nim
```diff
@@ -0,0 +1,176 @@
...
+    (response, updatedRef) = self.request(
+      keyRef,
+      auth       = self.writeAuth,
+      httpMethod = HttpPut,
+      body       = data,
+    )
+  if response.code.is2xx():
+    return updatedRef
+  else:
+    raise newException(ValueError, $response.code & " " & response.body())
```
mbaltrusitis, line 168: Object store API URLs can be finicky when initially configuring. Additional logging to handle 4xx errors could improve user experience.

src/object_store/api.nim
```diff
@@ -0,0 +1,148 @@
...
+        storeRef = storeConfig.store.objectExists(storeConfig, lookupRef)
+        if storeRef == nil:
+          error("object store: could not upload object " & $lookupRef & " due to: " & e)
+          dumpExOnDebug()
+  except:
+    error("object store: could not lookup existing object " & $lookupRef & " due to: " & getCurrentExceptionMsg())
+    dumpExOnDebug()
+  if storeRef != nil:
+    objectRefs[k] = storeRef
+    dict[objectStorePrefix & k] = pack($storeRef)
```
mbaltrusitis, line 120: should this be the `else` to balance the `if` on line 98?

mbaltrusitis (commented):
Overall looks good! I have a few suggestions around logging and the handling of the race condition which can could be made a bit clearer.

src/object_store/api.nim
miki725: ooh good catch. yeah forgot to clean up

src/object_store/api.nim
miki725: no

this is reference the var from:

https://github.com/crashappsec/chalk/blob/962a9b6698f2107234058746b6284a9027faec88/src/object_store/api.nim#L94

then inside the try statement various places attempt to set it if possible:

* query existing object
* if fails, attempt to upload object
* if fails attempt to query existing object again to handle concurrent upload race conditions

then here if any of them succeeded, add the reference to the `dict`. its not related to the if statements which define the fallback procedures for setting the `storeRef` in general

miki725:
> handling of the race condition which can could be made a bit clearer.

can you elaborate?

src/object_store/api.nim
mbaltrusitis: understood. thx!

mbaltrusitis:
you cleared up my understanding in your previous comment :+1:

mbaltrusitis (approved):

PR #496, opened 2025-03-04
docs: initial directory structure

mbaltrusitis:
validate all links in docs:

```
➜ markdown-link-check docs/**.md -q
```

PR #512, opened 2025-05-14
docs: update configs for chalkproject.io [ci skip]

miki725:
some of these were not merged as part of https://github.com/crashappsec/chalk/pull/496

also cleaning up some of the older ones which we do not reference anywhere and will prolly need to be refactored after n00b conversion so removing them altogether right now


MyNameIsMeerkat (approved):
Shippit

PR #528, opened 2025-06-19
refactor: switch to openssl buf2hexstr()

miki725:
old implementation wasnt complicated but might as well use openssl utils...


miki725:
probably simpler to keep our implementation with a simple loop

PR #541, opened 2025-07-11
fix: parse encoded build observables data

nettrino:


miki725 (changes requested):
Return type doesn't match now. It's returned as a string in the place but a dict in another

Also is it a binary string?

nettrino:
I want to return the contents of the file as returned by `read_file` which I thought returns binary? Isn't 
```
`x 
```

the right type for binary? how should this be declared for the value to just be binary data?

miki725 (changes requested):
there is no return type in the negative case now which will throw errors

```
`x
```

its a generic type. it allows the return type to be anything.

chalk doesnt have a binary type for collected data. it should be a valid string which can be reportable in json for example when object store fails

if we want to change to actual binary data that will require chalk and probably con4m changes

miki725 (commented):
can you adjust `type: string` above. cant comment it as a suggestion on the PR as its above the diff

PR #546, opened 2025-07-15
feat(dust): dust extension repo setup

mbaltrusitis:

## Description

Setting up dust sub-project as we approach our initial release. Release automation and docs to follow.

### DIFF
--- a/dust-lambda-extension/Makefile
+++ b/dust-lambda-extension/Makefile
@@ -13,6 +13,10 @@ IS_DIRTY := $(shell git diff-index --quiet HEAD -- 2>/dev/null || echo "dirty")
 SOURCES := $(wildcard extensions/*)
 BUILD_PREFIX ?= "."
 BUILD_DIR ?= $(BUILD_PREFIX)/_build
+AWS_LAMBDA_RUNTIME_API?=localhost:8585
+TASK_PATH?=..
+export AWS_LAMBDA_RUNTIME_API
+export TASK_PATH
 
 # Determine archive name based on git status
 ifeq ($(TAG),)
@@ -36,6 +40,9 @@ $(ARCHIVE_NAME): $(SOURCES)
 
 deploy-layer: ## TODO
 	@echo "Deploying..."
+
+test: ## test dust script against fake lambda runtime
+	./extensions/dust
 
 #> UTILITIES
 
--- a/dust-lambda-extension/Makefile
+++ b/dust-lambda-extension/Makefile
@@ -1,0 +1,90 @@
+ifneq (,)
+.error This Makefile requires GNU Make.
+endif
+
+# makefile config
+.ONESHELL:
+SHELL := /bin/bash
+.SHELLFLAGS := -eu -o pipefail -c
+MAKEFLAGS += --warn-undefined-variables
+
+# user config
+PYTHON3 ?= python3.12
+VENV_DIR := .venv-dust-lambda-extension
+VENV_PYTHON3 := ${VENV_DIR}/bin/python3
+
+#> ENVIRONMENT SETUP
+init: venv pre-commit-configured ## initialize the development environment
+.PHONY: init
+
+pre-commit-configured: pre-commit-installed ## configure pre-commit hooks
+	source $(VENV_DIR)/bin/activate
+	$(VENV_DIR)/bin/pre-commit install
+.PHONY: pre-commit-configured
+
+pre-commit-installed: $(VENV_DIR)/ ## install pre-commit in venv
+	$(VENV_DIR)/bin/pip install pre-commit
+.PHONY: pre-commit-installed
+
+venv: $(VENV_DIR) ## create Python virtualenv
+.PHONY: venv
+
+$(VENV_DIR):
+	@echo "Creating venv with $(PYTHON3)"
+	$(PYTHON3) -m venv $@
+
+#> LINTING
+lint:  ## run bash and markdown linters
+	pre-commit run --all-files
+.PHONY: lint
+
+#> DEPLOYMENT
+package-layer: ## create a zip archive for AWS Lambda Layer deployment
+	@echo "Creating Lambda Layer package..."
+	# Determine archive name based on git status and tags
+	@if git diff-index --quiet HEAD --; then \
+		if git describe --exact-match HEAD >/dev/null 2>&1; then \
+			TAG=$$(git describe --exact-match HEAD 2>/dev/null); \
+			# if commit is tagged and worktree is clean, use the tag
+			ARCHIVE_NAME="dust-lambda-extension-$$TAG.zip"; \
+		else \
+			SHORT_HASH=$$(git rev-parse --short HEAD); \
+			# if commit is untagged and worktree is clean, use short hash
+			ARCHIVE_NAME="dust-lambda-extension-$$SHORT_HASH.zip"; \
+		fi; \
+	else \
+		SHORT_HASH=$$(git rev-parse --short HEAD); \
+		# if commit is untagged and worktree is dirty, use {short hash}-dirty
+		ARCHIVE_NAME="dust-lambda-extension-$$SHORT_HASH-dirty.zip"; \
+	fi; \
+	@echo "Creating archive: $$ARCHIVE_NAME"; \
+	zip -r "$$ARCHIVE_NAME" extensions/; \
+	@echo "Lambda Layer package created: $$ARCHIVE_NAME"
+.PHONY: package-layer
+
+deploy-layer: ## TODO
+	@echo "Deploying..."
+
+#> UTILS
+clean: ## delete Lambda layer zip archives
+	rm -f dust-lambda-extension-*.zip
+.PHONY: clean
+
+clean-all: clean ## remove zip archives and development environment
+	rm -fr $(VENV_DIR)
+.PHONY: clean-all
+
+help: ## show this help
+	@echo "Specify a command from the list below:"
+	@echo ""
+	@awk 'BEGIN {FS = ":.*?## "}; \
+		/^#> / { \
+			gsub(/^#> /, "", $$0); \
+			printf "\n\n\033[1;35m%s\033[m\n", $$0; \
+			next \
+		} \
+		/^[0-9a-zA-Z_-]+:.*?## / { \
+			printf "  \033[0;36m%-25s\033[m %s\n", $$1, $$2 \
+		}' $(MAKEFILE_LIST)
+	@echo ""
+.PHONY: help
--- a/dust-lambda-extension/Makefile
+++ b/dust-lambda-extension/Makefile
@@ -3,76 +3,45 @@ ifneq (,)
 endif
 
 # makefile config
-.ONESHELL:
-SHELL := /bin/bash
-.SHELLFLAGS := -eu -o pipefail -c
 MAKEFLAGS += --warn-undefined-variables
 
-# user config
-PYTHON3 ?= python3.12
-VENV_DIR := .venv-dust-lambda-extension
-VENV_PYTHON3 := ${VENV_DIR}/bin/python3
-
-#> ENVIRONMENT SETUP
-init: venv pre-commit-configured ## initialize the development environment
-.PHONY: init
-
-pre-commit-configured: pre-commit-installed ## configure pre-commit hooks
-	source $(VENV_DIR)/bin/activate
-	$(VENV_DIR)/bin/pre-commit install
-.PHONY: pre-commit-configured
-
-pre-commit-installed: $(VENV_DIR)/ ## install pre-commit in venv
-	$(VENV_DIR)/bin/pip install pre-commit
-.PHONY: pre-commit-installed
-
-venv: $(VENV_DIR) ## create Python virtualenv
-.PHONY: venv
-
-$(VENV_DIR):
-	@echo "Creating venv with $(PYTHON3)"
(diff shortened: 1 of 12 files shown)


dust-lambda-extension/extensions/dust
```diff
@@ -0,0 +1,294 @@
...
+
+# Enable debug mode only if DEBUG environment variable is set
+[[ ${DEBUG:-} == "true" ]] && set -x
+
+# Logger helpers with consistent formatting and timestamps
+log_with_level() {
+	local level="$1"
+	local color="$2"
+	local message="$3"
+	printf "\e[%sm[%s] [%s] %s\e[0m\n" "$color" "$(date -Iseconds)" "$level" "$message" >&2
```
miki725, line 19: nit. `printf` I think is harder to read vs simple `echo`

for example in setup chalk:

https://github.com/crashappsec/setup-chalk-action/blob/44e40964dfd031b5d6b77ce5015a418c673ff649/setup.sh#L5-L50

dust-lambda-extension/extensions/dust
```diff
@@ -0,0 +1,294 @@
...
+LOG_WARNING() { log_with_level "WARN" "0;33" "$1"; }
+
+# Configurables (set via environment variables)
+readonly EXTENSION_NAME="${EXTENSION_NAME:-dust}"
+readonly MAX_FAILURES=${MAX_FAILURES:-5}
+
+# Constants
+readonly CHALK_BIN_PATH="$TASK_PATH/chalk"
+readonly CHALK_MARK_PATH="$TASK_PATH/chalk.json"
+readonly TASK_PATH="/var/task"
```
miki725, line 34: `TAS_PATH` is refed before its defined

dust-lambda-extension/extensions/dust
```diff
@@ -0,0 +1,294 @@
+#!/usr/bin/env bash
```
miki725, line 1: does this need to be a bash script? what does it do which actually requires bash?

can this rely simply on `/bin/sh`

lambda runtimes are used in other envs such as https://www.localstack.cloud/ which might not contain `bash`

unless we absolutely we need `bash`, lets not rely on it

dust-lambda-extension/extensions/dust
```diff
@@ -0,0 +1,294 @@
...
+
+# Configurables (set via environment variables)
+readonly EXTENSION_NAME="${EXTENSION_NAME:-dust}"
+readonly MAX_FAILURES=${MAX_FAILURES:-5}
+
+# Constants
+readonly CHALK_BIN_PATH="$TASK_PATH/chalk"
+readonly CHALK_MARK_PATH="$TASK_PATH/chalk.json"
+readonly TASK_PATH="/var/task"
+readonly TMP_CHALK_PATH="/tmp/chalk"
```
miki725, line 35: usually good practice to generate random temporary filename via `mktemp` unless we need a fixed path here

dust-lambda-extension/extensions/dust
```diff
@@ -0,0 +1,294 @@
...
+
+# Validate critical environment variable
+if [[ -z ${AWS_LAMBDA_RUNTIME_API:-} ]]; then
+	LOG_ERROR "AWS_LAMBDA_RUNTIME_API environment variable is not set"
+	exit 1
```
miki725, line 40: common pattern to have a util function like `fatal` which logs an error and then `exit 1`:

https://github.com/crashappsec/setup-chalk-action/blob/44e40964dfd031b5d6b77ce5015a418c673ff649/setup.sh#L46-L49

dust-lambda-extension/extensions/dust
```diff
@@ -0,0 +1,294 @@
...
+##
+
+set -euo pipefail
+
+# Enable debug mode only if DEBUG environment variable is set
+[[ ${DEBUG:-} == "true" ]] && set -x
+
+# Logger helpers with consistent formatting and timestamps
+log_with_level() {
+	local level="$1"
```
miki725, line 16: tabs in shell 😭 

lets match formatting in setup-chalk-action

https://github.com/crashappsec/setup-chalk-action/blob/main/.editorconfig

shfmt honors editorconfigs so we can add that

we can copy some of the nim configs from https://github.com/crashappsec/chalk-internal/blob/main/.editorconfig to make it a global config for the repo

dust-lambda-extension/extensions/dust
```diff
@@ -0,0 +1,294 @@
...
+	local method="$1"
+	local url="$2"
+	local headers=("${@:3}")
+
+	local temp_file
+	temp_file=$(mktemp)
+	local http_code
+
+	# Use array to properly handle headers with spaces
+	local curl_args=(-s -w "%{http_code}" -o "$temp_file" -X "$method" "$url")
```
miki725, line 63: this wont fail the command if anything goes bad

dust-lambda-extension/extensions/dust
```diff
@@ -0,0 +1,294 @@
...
+	if [[ $http_code -lt 200 || $http_code -ge 300 ]]; then
+		LOG_ERROR "HTTP request failed with status $http_code for $method $url"
+		LOG_DEBUG "Response: $(cat "$temp_file")"
+		rm -f "$temp_file"
+		return 1
+	fi
+
+	cat "$temp_file"
+	rm -f "$temp_file"
+	return 0
```
miki725, line 86: due to `-e` in the overall script, if anything fails the temp files wont be cleaned up

in chalk setup action this is handled gracefully:

https://github.com/crashappsec/setup-chalk-action/blob/44e40964dfd031b5d6b77ce5015a418c673ff649/setup.sh#L51-L65

dust-lambda-extension/extensions/dust
```diff
@@ -0,0 +1,294 @@
...
+		rm -f "$temp_file"
+		return 1
+	fi
+
+	if [[ $http_code -lt 200 || $http_code -ge 300 ]]; then
+		LOG_ERROR "HTTP request failed with status $http_code for $method $url"
+		LOG_DEBUG "Response: $(cat "$temp_file")"
+		rm -f "$temp_file"
+		return 1
+	fi
```
miki725, line 82: why not just use `-f` in curl with the `||` for error handling:

https://github.com/crashappsec/setup-chalk-action/blob/44e40964dfd031b5d6b77ce5015a418c673ff649/setup.sh#L201-L213

I think thats a lot more readable inline vs handling each error case individually

dust-lambda-extension/extensions/dust
```diff
@@ -0,0 +1,294 @@
...
+
+	return 0
+}
+
+# Process chalk marks with better error handling
+process_chalk_marks() {
+	LOG_INFO "$TMP_CHALK_PATH extracting $CHALK_MARK_PATH"
+	if [[ -f $CHALK_MARK_PATH ]]; then
+		LOG_DEBUG "Found $CHALK_MARK_PATH, running chalk extract."
+		if ! "$TMP_CHALK_PATH" extract "$CHALK_MARK_PATH"; then
```
miki725, line 143: why `extract` vs `env` here?

extract only extracts chalkmark from known codec artifacts which `chalk.json` is not as its not parsable by any codec

```
➜ cat chalk.json | jq '.METADATA_ID'
"5BX9CS-7XTG-5BJE-1N82ZR"

➜ chalk extract --log-level=none chalk.json
[
  {
    "_OPERATION": "extract",
   ...
    "_OP_CHALK_COUNT": 0,
    "_OP_UNMARKED_COUNT": 0
  }
]
```

dust-lambda-extension/extensions/dust
```diff
@@ -0,0 +1,294 @@
...
+	else
+		LOG_WARNING "No chalk.json found at $CHALK_MARK_PATH"
+		return 1
+	fi
+
+	return 0
+}
+
+# JSON parsing using pure bash/standard tools (no jq dependency)
+extract_json_field() {
```
miki725, line 156: lambda does not send these as return headers?

makes for parsing responses much much better

dust-lambda-extension/extensions/dust
```diff
@@ -0,0 +1,294 @@
...
+
+extract_extension_id() {
+	local temp_file="$1"
+	# Extract the Extension ID from response headers
+	EXTENSION_ID=$(awk '/^[Ll]ambda-[Ee]xtension-[Ii]dentifier:/ {
+        gsub(/\r$/, "");
+        gsub(/^[^:]*:[[:space:]]*/, "");
+        print $0;
+        exit
+    }' "$temp_file")
```
miki725, line 181: in chalk setup action

https://github.com/crashappsec/setup-chalk-action/blob/44e40964dfd031b5d6b77ce5015a418c673ff649/setup.sh#L170-L177

I think awk syntax is less obvious to people generally vs simple `grep`

as `curl` header output is known, this could be easily grepped

dust-lambda-extension/extensions/dust
```diff
@@ -0,0 +1,294 @@
...
+register_extension() {
+	# Only register for SHUTDOWN events (INVOKE events are not needed)
+	local registration_payload='{"events":["SHUTDOWN"]}'
+
+	# Make a single registration request with headers included
+	local temp_file
+	temp_file=$(mktemp)
+	local http_code
+
+	http_code=$(curl --silent --write-out "%{http_code}" \
```
miki725, line 200: why not use the curl wrapper above?

dust-lambda-extension/extensions/dust
```diff
@@ -0,0 +1,294 @@
...
+		--dump-header "$temp_file" \
+		--request POST "$REGISTER_API" \
+		--header "Lambda-Extension-Name: $EXTENSION_NAME" \
+		--header "Content-Type: application/json" \
+		--data "$registration_payload" \
+		--output /dev/null)
+
+	if [[ $http_code -ne 200 ]]; then
```
miki725, line 208: similar error handoling concerns as above.

also as this use `-w` for the output format, if the response does include any useful error message, it wont be showed here as this will only print headers (where there might not be any)

see how setup chalk action does it

dust-lambda-extension/extensions/dust
```diff
@@ -0,0 +1,294 @@
...
+			fi
+
+			# Brief backoff before retrying
+			sleep 1
+		fi
+	done
+}
+
+# Run main function if script is executed directly
+if [[ ${BASH_SOURCE[0]} == "${0}" ]]; then
```
miki725, line 292: do we expect people to source this script?

checking `BASH_SOURCE` is needed to separate `source foo.sh` vs `./foo.sh` but if its not a concern here usually shell scripts just execute directly whatever they need

dust-lambda-extension/extensions/dust
```diff
@@ -0,0 +1,294 @@
...
+	fi
+
+	# Process chalk marks
+	process_chalk_marks || LOG_WARNING "Chalk mark processing failed, continuing anyway"
+
+	# Poll for events from the Lambda Runtime API
+	LOG_INFO "Starting event polling loop."
+
+	local consecutive_failures=0
+	local max_failures=$MAX_FAILURES
```
miki725, line 242: why alias local vars to exactly copy of a global var?

dust-lambda-extension/README.md
```diff
@@ -0,0 +1,51 @@
...
+## getting started.
+
+A [Makefile](./Makefile) is provided for convenience to perform common setup and
+development operations, including bootstrapping your development environment.
+After satisfying the [dependencies](#dependencies) below, simply run
+`make init`.
+
+**dependencies:**
+
+- Bash >=5.1
```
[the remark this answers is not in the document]
mbaltrusitis: its the only one i tested it with

dust-lambda-extension/Makefile
```diff
@@ -0,0 +1,90 @@
+ifneq (,)
+.error This Makefile requires GNU Make.
+endif
+
+# makefile config
+.ONESHELL:
+SHELL := /bin/bash
+.SHELLFLAGS := -eu -o pipefail -c
```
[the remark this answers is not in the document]
mbaltrusitis: this was copied over from some boilerplate. i'll adjust.

dust-lambda-extension/Makefile
```diff
@@ -0,0 +1,90 @@
...
+.PHONY: venv
+
+$(VENV_DIR):
+	@echo "Creating venv with $(PYTHON3)"
+	$(PYTHON3) -m venv $@
+
+#> LINTING
+lint:  ## run bash and markdown linters
+	pre-commit run --all-files
+.PHONY: lint
```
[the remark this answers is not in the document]
mbaltrusitis: done :heavy_check_mark:

dust-lambda-extension/extensions/dust
mbaltrusitis: fixed

dust-lambda-extension/extensions/dust
mbaltrusitis: It returns the event UUID but not the event type

dust-lambda-extension/extensions/dust
mbaltrusitis: This extracts headers, the wrapper extracts the body.

dust-lambda-extension/extensions/dust
mbaltrusitis: I typically take a local copy for functions to keep state encapsulated. I'll dedupe.

miki725 (approved):

PR #563, opened 2025-08-21
fix: dont inotify system folders like /proc

miki725:
k8s seems to be mounting some files in /proc so we just ignore them

the system folders are grabbed from nimutils

## Issue

chalk has errors like:

```
error: inotify: could not watch /proc/sysrq-trigger
```

## Description

ignore system folders from using inotify

## Testing

test with k8s container

### DIFF
--- a/src/utils/files.nim
+++ b/src/utils/files.nim
@@ -22,6 +22,7 @@ import ".."/[
 import "."/[
   fd_cache,
   file_string_stream,
+  strings,
   times, # TODO remove
 ]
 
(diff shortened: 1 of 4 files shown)


src/utils/files.nim
```diff
@@ -168,7 +169,7 @@ when defined(linux):
         if len(parts) <= 2:
           continue
         let path = parts[1]
-        if path.fileExists():
+        if not path.startsWithAnyOf(systemIgnoreStartsWithPaths) and path.fileExists():
```
miki725, line 172: fyi defined here https://github.com/crashappsec/nimutils/blob/ada76cf776b9506e6efff32fffdaa58c65a9a2cf/nimutils/file.nim#L531-L540

orkhanM (approved):

PR #568, opened 2025-08-28
fix: handle confspec errors during chalk load

miki725:

## Issue

during chalk load which is registering additional keys would see:

```
error: When collecting chalk-time host info, plugin implementation conffile threw an exception it didn't handle: get() of an attribute that isn't set.
error: When collecting chalk-time artifact data, plugin implementation conffile threw an exception it didn't handle (artifact = /home/root/Code/co/setup-chalk-action/chalk): get() of an attribute that isn't set.
error: When collecting run-time host info, plugin implementation conffile threw an exception it didn't handle: get() of an attribute that isn't set.
```

## Description

while testing the configuration some keys are partially registered which shows non-errors as error logs while calculating the report while doing the command report

## Testing

for the log updates:

```
➜ ./chalk load configs/debug.c4m
```

### DIFF
--- a/src/plugins/conffile.nim
+++ b/src/plugins/conffile.nim
@@ -26,7 +26,7 @@ proc scanForWork(kt: auto, opt: Option[ChalkObj], args: seq[Box]): ChalkDict =
     try:
       # during configuration testing, some keys are registered
       # which are not fully initialized so just skip here
-      # if the kind id missing.
+      # if the "kind" id missing.
       # otherwise this should never happen as the config is validated
       # while loading
       if attrGet[int](v & ".kind") != int(kt):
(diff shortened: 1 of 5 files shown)


src/plugins/conffile.nim
```diff
@@ -19,14 +19,26 @@ proc scanForWork(kt: auto, opt: Option[ChalkObj], args: seq[Box]): ChalkDict =
...
-    if attrGet[int](v & ".kind") != int(kt): continue
-    if not isSubscribedKey(k):
+    if opt.isNone() and k in hostInfo:
+      continue
+    if opt.isSome() and k in opt.get().collectedData:
+      continue
+    try:
+      # during configuration testing, some keys are registered
+      # which are not fully initialized so just skip here
+      # if the kind id missing.
```
mbaltrusitis, line 29:
```suggestion
      # if the "kind" id missing.
```

mbaltrusitis (approved):

PR #657, opened 2026-04-26
feat: native Mach-O codec (LC_NOTE, no script wrapper)

viega:
## Summary

- Native macOS codec: marks Mach-O binaries via `LC_NOTE` (data_owner = `chalk`) with the chalk-mark JSON inside `__LINKEDIT`. Apple's notary accepts the result; `codesign --verify --strict` passes; `dyld` ignores our addition; `strip` preserves it.
- The script-wrapper codec (`codecMacOs.nim`) stays as a fallback at priority 2.
- Cross-platform: chalk on Linux can read both wrapper and native marks on Mac binaries, and write native marks on unsigned Mach-Os.

## Verification

- chalk built natively on macOS arm64 and inside Docker (alpine). The chalk binary is itself a real Mach-O after `make release`.
- **Apple's notary `Accepted`** both the binary tarball and the `.pkg` installer produced by `scripts/release-macos.sh`.
- Linux chalk extracts marks written by macOS chalk and vice versa; cross-platform interop confirmed.
- Unit test (`tests/unit/test_macho.nim`) covers parse / signature classification / mark+unmark round-trip / unchalked-hash invariance. Picked up by `nimble test` and the existing unit-tests CI workflow.
- 0 leaks under macOS `leaks(1)`.

| Codec path | Read | Write | Re-sign |
|---|---|---|---|
| Native LC_NOTE on macOS | ✓ | ✓ | ad-hoc auto / DevID via release-macos.sh |
| Native LC_NOTE on Linux | ✓ | ✓ (skips re-sign w/ warn) | n/a |
| Wrapper script on macOS | ✓ | ✓ | n/a |
| Wrapper script on Linux | ✓ | ✓ | n/a |
| Dev-ID-signed Mach-O | ✓ extract | refuse w/ warn | n/a |

## Bundled fix

\`src/commands/cmd_exec.nim\`'s \`doPostExec\` body uses inotify primitives unconditionally; gated under \`when hostOS == \"linux\":\` so the macOS build can compile. The runtime was already a no-op on non-Linux (\`initPostExecWatch\` returns none); only the type-check needed gating. Pre-existing chalk bug, surfaced by my macOS-host build.

## Reference

[\`docs/design-macho-codec.md\`](docs/design-macho-codec.md) covers what's stored, what's refused, the mutation flow, and the unchalked-hash semantics.

### What you can run as a reviewer

- [ ] CI builds + unit-tests pass (in this PR)
- [ ] \`make tests\` locally (functional suite)
- [ ] \`make DOCKER= release\` on macOS produces a real Mach-O \`./chalk\` (not a script wrapper). Reviewer doesn't need a Developer ID cert for this — \`make release\` self-marks via the codec and re-applies an ad-hoc signature, which is enough to run locally.
  - Native macOS build env requirements (not yet documented in-repo; happy to land a docs/build-macos.md if useful):
    - \`SDKROOT=$(xcrun --show-sdk-path)\` exported.
    - \`/opt/homebrew/opt/openssl@3/bin\` ahead of \`/usr/bin\` in PATH (LibreSSL ships without \`storeutl\`, which nimutils' compile-time CA-bundle validation needs).
    - \`libsodium.a\` staged into \`~/.local/c0/libs/\` from \`brew --prefix libsodium\`.

### What you can NOT run as a reviewer

- [ ] **\`scripts/release-macos.sh\` (notarized, signed, stapled \`.pkg\` + tarball)** — only @viega can run this today. The pipeline needs CO's Developer ID certs and the CRAYON_NOTARY notarytool keychain-profile, all bound to John's login keychain. Anyone else attempting it will fail at codesign (no cert) or at notary submit (CRAYON_NOTARY profile not present in their keychain).
- [ ] **Cutting an actual release** — only @viega can do that today. The git tag, the version bump, and the publish-to-GitHub-releases all flow through John.

### Path to CI/CD-minted macOS releases

To unblock this from being John-bottlenecked, a follow-up PR will need:

1. **Crash Override provisions a dedicated CI/CD identity** with Apple — separate from John's personal Developer ID profile:
   - A new \`Developer ID Application\` certificate issued under the CO team account (F9V4P5LT8S).
   - A new \`Developer ID Installer\` certificate, same team.
   - A new app-specific password + Apple ID (or App Store Connect API key) for notarytool, also under the CO team. Cleanest: an App Store Connect API key (\`AuthKey_*.p8\` + key ID + issuer ID), since those don't require an Apple ID password rotation cadence and aren't bound to a human's account.
2. **A self-hosted macOS arm64 runner** (or GitHub-hosted \`macos-14\`+) with Xcode CLT installed.
3. **GitHub Actions secrets** holding the CI identity:
   - \`APPLE_CERT_APPLICATION_P12\` / \`APPLE_CERT_INSTALLER_P12\` (base64'd \`.p12\` exports) plus their passwords.
   - \`APPLE_NOTARY_API_KEY_ID\` / \`APPLE_NOTARY_API_ISSUER_ID\` / \`APPLE_NOTARY_API_KEY_P8\` (the App Store Connect key triple).
4. **Job body**: spin up an ephemeral keychain at job start, \`security import\` the certs, point notarytool at the API key, then run essentially the existing \`scripts/release-macos.sh\` (with \`APPLICATION_IDENTITY\`/\`INSTALLER_IDENTITY\`/\`NOTARY_PROFILE\` overridden to the CI identity, and the env-overrides comment block in the script extended accordingly).
5. **Trigger**: \`workflow_dispatch\` (manual mint) and/or \`on: release\` (when a tag like \`v1.0.1\` is published from \`main\`).

Out of scope for this PR — but unblocked by it. Without the native codec, none of this would matter (the script-wrapper output isn't notarizable).


src/commands/cmd_exec.nim
```diff
@@ -280,95 +280,103 @@ proc doPostExec(state: Option[PostExecState], detach: bool) =
...
+  # from type-checking inotify primitives on other platforms.
+  when hostOS != "linux":
+    return
 
-  let
-    ws            = state.get()
-    niceValue     = attrGet[int]("exec.postexec.nice")
-  var
-    accessedPaths = initHashSet[string]()
+  else:
```
miki725:
```suggestion
  when hostOS == "linux":
```

should do the same trick

src/plugins/codecMacho.nim
```diff
@@ -0,0 +1,294 @@
...
+    sigKind:   ChalkMachoSigKind
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+proc adhocResign(path: string): bool =
+  ## Restore an ad-hoc signature on a binary we just mutated.
+  ## Returns true on success.  Emits a warn on failure.
+  if findExe("codesign") == "":
```
miki725: I think there were edge cases around nimd `findExe` before

we mostly use `findExePath` in chalk

https://github.com/crashappsec/chalk/blob/07f88aa3d5e42d85ff6579afd2f1b204c9bdafb4/src/utils/exe.nim#L27-L31

it also allows to ignore chalk binaries in case the exe is wrapped like `docker`

but prolly this is good for now

src/plugins/codecMacho.nim
```diff
@@ -0,0 +1,294 @@
...
+    warn("codesign not found in PATH — cannot re-sign " & path &
+         " (was ad-hoc-signed before chalking)")
+    return false
+
+  let (output, exitCode) = execCmdEx("codesign --force --sign - " &
+                                      quoteShell(path))
```
miki725: I think there were issues with this before hence you wrote all the switchboard stuff?

rest of chalk uses the switchboard for the subproc stuff now

https://github.com/crashappsec/chalk/blob/07f88aa3d5e42d85ff6579afd2f1b204c9bdafb4/src/docker/exe.nim#L54

src/plugins/codecMacho.nim
```diff
@@ -0,0 +1,294 @@
...
+         "treating as unmarked")
+
+  let chalk = newChalk(
+    name         = path,
+    fsRef        = path,
+    codec        = self,
+    resourceType = {ResourceFile},
+    cache        = cache,
+    extract      = dict,
+    marked       = marked,
```
miki725: nit

```suggestion
```

this is already handled in `newChalk` and it actually only puts it as marked when there is any metadata around of len > 0

src/plugins/codecMacho.nim
```diff
@@ -0,0 +1,294 @@
...
+# Registration
+# ---------------------------------------------------------------------------
+
+proc loadCodecMacho*() =
+  newCodec(
+    "macho",
+    # Listed as native on macOS so `chalk load <config>` self-marks
+    # via this codec when run on a Mac.  Also native on Linux because
+    # chalk on Linux can usefully mark cross-compiled Mac binaries.
+    nativeObjPlatforms = @["macosx", "linux"],
```
miki725: this populates native codecs which powers:

* exe detection when it needs to skip `chalk` itself such as for `docker`
* during self-copy which is used for create a docker binary to be embedded when entrypoint wrapping

by allowing both os here that will for example allow to embed mac chalk binary in a linux docker build as it will detect it as a valid chalk binary

for linux support either way if chalk can parse the file it can cross-mark it regardless of their codec

```suggestion
    # Listed as native on macOS to allow `chalk load <config>` self-marks
    # via this codec when run on a Mac
    nativeObjPlatforms = @["macosx"],
```

src/utils/macho.nim
```diff
@@ -0,0 +1,366 @@
...
+  ## SHA-256 hex of the canonicalized binary (32-byte zero chalk
+  ## payload form, mirroring elf.nim's getUnchalkedHash).  Returns ""
+  ## on error.
+  if self == nil or self.bin == nil:
+    return ""
+
+  var hex = newString(64)
+  # The C side writes 64 hex chars + NUL into out_hex; pre-size the
+  # string so cstring(hex) points at writable storage of >=65 bytes.
+  hex.setLen(65)
```
miki725: is this intentional?

not following this. `newString` should allocate nims string for 64 bytes plus a null-byte for null termination. But below the length is overwritten to a greater value of 65. cant that buffer overflow.

from what Ive seen before `newString` can allocate bigger buffer and then `setLen` is used to set exact length for example when converting back from non-null terminal strings and you dont know the length ahead of time

src/utils/macho.nim
```diff
@@ -0,0 +1,366 @@
...
+    var count: csize_t = 0
+    let arr = chalk_macho_get_notes(self.bin, addr count)
+    if arr != nil:
+      try:
+        for i in 0 ..< count.int:
+          let n = cast[ptr UncheckedArray[ChalkMachoNote]](arr)[i]
+          var owner = newString(0)
+          for c in n.data_owner:
+            if c == '\0': break
+            owner.add(c)
```
miki725, line 356: is this expected to be long?

we had similar problems in json encoding in nimutils before where it was appending a few chars at a time however internally nim ends up allocating new strings and then copying the data over which made the whole thing slow

once we switched to string stream that solved that problem. in that case it just expands the stream buffer without needing to realloc all the time

https://github.com/crashappsec/nimutils/blob/d4a1693591f9083027d66fd00c7d992d28070e84/nimutils/box.nim#L353-L358

prolly not a concern for small strings but if can be in Mbs might be better to use similar trick

docs/design-macho-codec.md
```diff
@@ -0,0 +1,110 @@
...
+## Cross-platform
+
+The C library has no macOS-only dependencies (libcrypto for
+SHA-256, plain POSIX for file I/O). The codec is registered as
+native on both macOS and Linux, so chalk on Linux can:
+
+- Read both wrapper and native chalk marks on Mac binaries.
+- Write native marks on unsigned Mach-Os (warns about no
+  `codesign` in PATH; binary won't run on macOS without re-signing
+  there).
```
miki725: chalk can do that regardless of the native codec registration. see another comment about that

scripts/release-macos.sh
miki725, line 1: would this need to be integrated in our mac release process?

miki725:
hmm something borked my mac binary:

copied from mac machine ls binary to linux:

```
➜ uname -a
Darwin Miroslavs-Mac-Studio.local 25.4.0 Darwin Kernel Version 25.4.0: Thu Mar 19 19:33:25 PDT 2026; root:xnu-12377.101.15~1/RELEASE_ARM64_T6041 arm64 arm Darwin

➜ scp (which ls) earth:/home/root/Code/co/chalk/macosls
```

then on linux machine chalked that binary:

```
➜ uname -a
Linux earth 6.19.10-arch1-1 #1 SMP PREEMPT_DYNAMIC Thu, 26 Mar 2026 19:20:28 +0000 x86_64 GNU/Linux

➜ ./chalk insert macosls
info:  /etc/chalk.c4m: Found config file
info:  attestation: signing file /home/root/Code/co/chalk/macosls
warn:  codesign not found in PATH — cannot re-sign /home/root/Code/co/chalk/macosls (was ad-hoc-signed before chalking)
info:  /home/root/.local/chalk/chalk.log: Open (sink conf='default_out')
info:  Full chalk report appended to: ~/.local/chalk/chalk.log
[
  {
    "_OPERATION": "insert",
    "_DATETIME": "2026-04-30T19:46:52.597-04:00",
    "_CHALKS": [
      {
        "CHALK_ID": "CMSP2S-HR60-R64R-B571HK",
        "PRE_CHALK_HASH": "1b649bab6267d94968f96dd3586fc692b0be587d409c51ea77e1d0be545135e2",
        "PATH_WHEN_CHALKED": "/home/root/Code/co/chalk/macosls",
        "ARTIFACT_TYPE": "Mach-O executable",
        "ORIGIN_URI": "git@github.com:crashappsec/chalk.git",
        "COMMIT_ID": "95b15740024043b3de6d0cc04f369d30cea4b873",
        "COMMIT_SIGNED": false,
        "BRANCH": "jtv/macho-native-codec",
        "AUTHOR": "Lee T.H. <l33th4@h4x0r.org>",
        "COMMITTER": "Lee T.H. <l33th4@h4x0r.org>",
        "COMMIT_MESSAGE": "perf(macho): magic-check before slurping full binary\n\nmachoScan previously called readFile(path) before checking the 4-byte\nMach-O magic, allocating and copying every scanned file end-to-end\njust to discard the bytes for non-Mach-O paths.  Mirror codecElf's\nwithFileStream + peek pattern so non-Mach-O files cost one open and\na 4-byte read.  Drop the dead MachoCache.rawBytes field while here.\n\nCo-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>",
        "DATE_AUTHORED": "2026-04-30T17:40:39.000-04:00",
        "TIMESTAMP_AUTHORED": 1777585239000,
        "DATE_COMMITTED": "2026-04-30T17:40:39.000-04:00",
        "TIMESTAMP_COMMITTED": 1777585239000,
        "CHALK_VERSION": "1.0.0",
        "METADATA_ID": "4J2JSH-G82J-NMS6-YQVA36",
        "SIGNATURE": "MEYCIQD7hmsA6btC+9UTQnhp6g1xWFmVfCkbcFsAph7sqIQWVAIhAJXrgqEgtSFAUBPVnpX977gHCkrdaa9+fowaq4PL1ZYy",
        "_VIRTUAL": false,
        "_CURRENT_HASH": "1b649bab6267d94968f96dd3586fc692b0be587d409c51ea77e1d0be545135e2"
      }
    ],
    "_AUTHOR": "Lee T.H. <l33th4@h4x0r.org>",
    "_BRANCH": "jtv/macho-native-codec",
    "_COMMITTER": "Lee T.H. <l33th4@h4x0r.org>",
    "_COMMIT_ID": "95b15740024043b3de6d0cc04f369d30cea4b873",
    "_COMMIT_MESSAGE": "perf(macho): magic-check before slurping full binary\n\nmachoScan previously called readFile(path) before checking the 4-byte\nMach-O magic, allocating and copying every scanned file end-to-end\njust to discard the bytes for non-Mach-O paths.  Mirror codecElf's\nwithFileStream + peek pattern so non-Mach-O files cost one open and\na 4-byte read.  Drop the dead MachoCache.rawBytes field while here.\n\nCo-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>",
    "_COMMIT_SIGNED": false,
    "_DATE_AUTHORED": "2026-04-30T17:40:39.000-04:00",
    "_DATE_COMMITTED": "2026-04-30T17:40:39.000-04:00",
    "_ENV": {
      "PATH": "/home/root/Code/co/n00b/bin:/home/root/Code/co/crash-argo-apps:/home/root/Code/co/chalk/tests/functional:/home/root/Code/co/chalk:/home/root/Code/co/util.sh/bin:/home/root/.bin:/home/root/.local/bin:/home/root/.nimble/bin:/home/root/.n/bin:/home/root/.go/bin:/home/root/.krew/bin:/home/root/.pyenv/versions/3.13.3/bin:/home/root/.pyenv/versions/3.12.10/bin:/home/root/.pyenv/versions/3.12.0/bin:/home/root/.pyenv/versions/3.11.6/bin:/home/root/.pyenv/versions/3.11.0/bin:/home/root/.pyenv/versions/3.10.9/bin:/home/root/.pyenv/versions/3.10.8/bin:/home/root/.pyenv/versions/3.9.16/bin:/home/root/.pyenv/versions/3.9.9/bin:/home/root/.pyenv/versions/3.9.7/bin:/home/root/.pyenv/versions/3.9.6/bin:/home/root/.pyenv/versions/3.9.5/bin:/home/root/.pyenv/versions/3.8.12/bin:/home/root/.pyenv/versions/3.8.11/bin:/home/root/.pyenv/versions/3.8.10/bin:/home/root/.pyenv/versions/3.7.12/bin:/home/root/.pyenv/versions/3.7.11/bin:/home/root/.pyenv/versions/3.7.10/bin:/home/root/.pyenv/versions/3.6.14/bin:/home/root/.pyenv/versions/2.7.18/bin:/usr/local/bin:/usr/local/sbin:/usr/bin:/bin:/usr/sbin:/sbin:/home/root/.fish-path-hook",
      "PWD": "/home/root/Code/co/chalk",
      "SSH_TTY": "/dev/pts/0",
      "USER": "root",
      "XDG_SESSION_TYPE": "tty",
      "AWS_SECRET_ACCESS_KEY": "<<redact>>"
    },
    "_OP_ANCESTOR_ARGVS": [
      [
        "fish",
        "--login"
      ],
      [
        "tmux"
      ],
      [
        "/usr/lib/systemd/systemd",
        "--switched-root",
        "--system",
        "--deserialize=50"
      ]
    ],
    "_OP_ARGV": [
      "/home/root/Code/co/chalk/chalk",
      "insert",
      "macosls"
    ],
    "_OP_CHALKER_VERSION": "1.0.0",
    "_ORIGIN_URI": "git@github.com:crashappsec/chalk.git",
    "_TIMESTAMP_AUTHORED": 1777585239000,
    "_TIMESTAMP_COMMITTED": 1777585239000,
    "_OP_CHALK_COUNT": 0,
    "_OP_UNMARKED_COUNT": 0
  }
]
```

the file staged as macho:

```
➜ file macosls
macosls: Mach-O 64-bit arm64 executable, flags:<NOUNDEFS|DYLDLINK|TWOLEVEL|PIE>
```

and I can it has chalkmark now:

```
➜ strings macosls  | grep MAGIC | jq | head -n2
{
  "MAGIC": "dadfedabbadabbed",
```

however when I copy it back to mac it blows up:

```
➜ scp  earth:/home/root/Code/co/chalk/macosls macosls
macosls

➜ ./macosls
fish: Job 1, './macosls' terminated by signal SIGKILL (Forced quit)
```

miki725:
the issue turned out to be when system integrity is enabled (its on by default on all macs) it does not allow to run non-signed binaries

that means that if linux chalks macho binary it will fail to run on any modern mac. well be changing so that if `codesign` is missing on linux, it will refuse to chalk stuff from linux but of course it can still extract from such binaries

src/utils/macho.nim
miki725: no its up to 17 chars

miki725 (approved):

PR #660, opened 2026-04-29
Provide chalk support for SHT_NOBITS section offsets beyond EOF

drraid:
Apparently some producers of legit ELFs will place section offsets beyond file EOF for SHT_NOBITS segment types. This PR moves one of the section offset checks into the next if-block, so that it only applies to section types which are not SHT_NOBITS. 

Testing was only performed on the `rclone` binary first observed exhibiting sections of type SHT_NOBITS with offsets beyond EOF.


miki725 (commented):
👍 🎉 

can we add an entry to the changelog that this was fixed

PR #669, opened 2026-05-27
feat: docker build context upload as OCI attestation

miki725:

## Summary

Chalk can now upload Docker build contexts as OCI attestations attached to pushed images. This allows downstream tools like Ocular to inspect the exact source files used to produce an image alongside the image itself — including any extra contexts passed via `--build-context name=path`.

## Configuration

Context upload is configured via a `docker_context_upload` singleton nested under `docker_push`:

```con4m
docker {
  docker_registry my_registry {
    uri:          "registry.example.com"
    login_method: "get"
    docker_login_get {
      uri: "https://auth.example.com/v1/docker-creds"
    }
    docker_push my_app {
      repository: "my-org/my-app"
      tags:       ["latest", "{BRANCH}"]
      docker_context_upload {
        enabled:  true
        strategy: "auto"
      }
    }
  }
}
```

### Key fields on `docker_context_upload`

| Field | Default | Description |
|---|---|---|
| `enabled` | `false` | Must be set to `true` to activate |
| `strategy` | `"auto"` | When/how the tarball is uploaded (see below) |
| `size_threshold` | `<<100mb>>` | Skip upload if total tarball exceeds this; reports `CONTEXT_TOO_LARGE` in `_OP_FAILED_KEYS` |
| `max_file_size` | `<<0mb>>` | Skip individual files exceeding this size (0 = no limit); skipped files reported in `_REPO_BUILD_CONTEXT_SKIPPED_FILES` |
| `additional_dockerignore` | `[".git"]` | Extra glob patterns appended after `.dockerignore` (last-match-wins) |
| `honor_dockerignore` | `true` | Apply `.dockerignore` patterns using Docker's own semantics |

### New `docker`-level fields

| Field | Default | Description |
|---|---|---|
| `build_context_cache_max_age` | `<<1 hrs>>` | Maximum age of cached tarballs under `/tmp/chalk-build-contexts/`; stale entries cleaned at both `docker build` and `docker push` |
| `registry_layer_chunk_size` | `<<5mb>>` | Chunk size for registry layer uploads |
| `registry_layer_upload_timeout` | `<<30 sec>>` | Timeout for a single chunk upload |

## Upload Strategies

Because `chalk docker build` and `chalk docker push` may run as separate commands, three strategies control when the tarball is created and uploaded:

| Strategy | Build time | Push time | Tarball lifetime |
|---|---|---|---|
| `registry` | Upload blob to registry | Create and push attestation manifest | No local file |
| `local` | Save `.tar.gz` to `/tmp/chalk-build-contexts/` | Upload blob + push manifest | Kept after push; deleted only by `build_context_cache_max_age` TTL |
| `disk` | Record context path in chalk mark | Read path, create tarball, upload, push manifest | No local file |
| `auto` | CI detected → `registry`; else → `local` | (as above) | (as above) |

CI is detected via environment variables: `CI`, `GITHUB_ACTIONS`, `GITLAB_CI`, `JENKINS_URL`, `CIRCLECI`, `TRAVIS`, `BUILDKITE`, `DRONE`, `SEMAPHORE`, `TEAMCITY_VERSION`, `BITBUCKET_BUILD_NUMBER`, `CODEBUILD_BUILD_ID`.

## OCI Representation

The context is stored as an OCI image manifest at the attestation tag:

```
<registry>/<repo>:sha256-<image-manifest-digest>
```

with:
- `artifactType: application/vnd.crashoverride.chalk.build-context.v1`
- Annotations: `org.opencontainers.image.created` and `dev.crashoverride.chalk.build-context.name`
- A `subject` referencing the image manifest it is attached to

Multiple attestation types (e.g. sigstore signatures and build contexts) coexist in a manifest list at the same tag; Chalk fetches the existing list before appending so nothing is overwritten.

## Context Filtering

When building the `.tar.gz`, Chalk applies patterns using Docker's `.dockerignore` semantics:

1. Chalk looks for the ignore file next to the Dockerfile: `<dockerfile-dir>/<dockerfile-basename>.dockerignore` (same priority order Docker uses — Dockerfile-specific file takes precedence over plain `.dockerignore` in the context root)
2. `additional_dockerignore` patterns are appended after the ignore-file patterns (last-match-wins)
3. Negation patterns (`!pattern`) re-include files inside excluded directories
4. Recursion into excluded directories is pruned unless a negation pattern could re-include files inside
5. Symlinks are stored as ustar typeflag-2 entries; long filenames use GNU LongLink preamble entries; dangling symlinks are archived as-is without following the link
6. Individual files exceeding `max_file_size` are skipped and their SHA-256 digest + size are reported in `_REPO_BUILD_CONTEXT_SKIPPED_FILES`
7. The size threshold is checked after every entry (directories, symlinks, regular files) so it fires regardless of context content

## New Chalk Keys

| Key | Kind | Description |
|---|---|---|
| `DOCKER_BUILD_CONTEXT_SNAPSHOTS` | chalk-time | Intermediate state embedded in the chalk mark, consumed at push time |
| `_REPO_BUILD_CONTEXTS` | runtime | Context manifest digests: `registry → repo → context name → digest` |
| `_REPO_BUILD_CONTEXT_TAR_SIZES` | runtime | Uploaded tarball sizes in bytes: `registry → repo → context name → bytes` |
| `_REPO_BUILD_CONTEXT_SKIPPED_FILES` | runtime | Files skipped due to `max_file_size`: `registry → repo → context name → path → {hash, size}` |

## Fixes Included

- OCI distribution spec `Content-Range` header format (`{first}-{last}`, no `bytes` prefix)
- `DOCKER_ADDITIONAL_CONTEXTS` chalk key now correctly captures `--build-context` named contexts
- `-std=c2x` used on macOS (Apple clang < 18 does not support `-std=c23`)
- `.dockerignore` lookup: look next to the Dockerfile (`<dockerfile-dir>/<dockerfile-basename>.dockerignore`), not in the context root
- Size threshold now fires after every entry type (directories, symlinks, EOF blocks), not only after regular file data
- Tar cache filename is collision-resistant: `<chalkId>-<configSlug>-<contextSlug>-<sha256(contextName)[0:8]>.tar.gz`
- Local strategy retains tarball after push so the same archive can be reused for pushes to multiple registries
- Local strategy avoids double-reading the tarball: pre-computed SHA-256 is passed into the manifest so the upload path does not rehash the file
- TOCTOU guard: after writing each file's body, the byte count written is compared against the stat'd size; a mismatch (file mutated mid-walk) errors the whole context rather than silently producing a corrupt tar
- Dangling symlinks no longer abort context upload: `lstat` (`followSymlink=false`) is used for mtime so broken symlinks don't raise during archiving
- `ChalkDict.merge` now propagates `deep=true` through recursion, fixing 3-level nested table merges (`registry → repo → context name`)
- Pre-1970 file modification times are clamped to `0` in ustar headers; positive overflow still raises `ValueError`
- `honor_dockerignore` default corrected to `true` when restoring disk-strategy snapshots (was inadvertently `false`)
- `globMatch` reimplemented as an iterative anchor-stack algorithm: O(k) space for k `**` segments (vs O(n×m) memoised recursion), matching the approach used in Go's glob implementations

See `docs/design-docker-registry.md` for the full design.

## Testing

```
make tests args="test_docker.py::test_build_context_upload -x"
make unit-tests args="pattern tests/unit/test_tar.nim"
```

Pattern matching is also validated against a shared JSON test suite at `tests/equivalency/dockerignore/testcases.json` run against both the Go `moby/patternmatcher` reference and the Nim implementation:

```
cd tests/equivalency/dockerignore && go test ./...
make unit-tests args="pattern tests/unit/test_tar.nim"
```


config.nims
```diff
@@ -16,7 +16,7 @@ var
   default  = getEnv("HOME").joinPath(".local/c0")
   localDir = getEnv("LOCAL_INSTALL_DIR", default)
   libDir   = localdir.joinPath("libs")
-  libs     = ["pcre", "ssl", "crypto", "gumbo", "hatrack", "sodium"]
+  libs     = ["pcre", "ssl", "crypto", "gumbo", "hatrack", "sodium", "z"]
```
viega, line 19: **🔴 Blocking until verified — static zlib link vs. Linux/Docker build.**

Adding `"z"` here feeds `staticLinkLibraries(libs, libDir, muslBase = localDir)`, so the musl/Alpine static build now requires `libz.a` in `$LOCAL_INSTALL_DIR/libs`. But the provisioning step in the `Dockerfile` is commented out (see other comment), so this links cleanly **only if** the pinned nimutils deps package already ships static zlib. Native macOS links system `libz`, so a local mac build won't catch a regression here. Please confirm `libz.a` is present in the deps the con4m/nimutils pin resolves to, or run the actual `docker compose run --rm chalk nimble release`.

Dockerfile
```diff
@@ -50,3 +50,10 @@ WORKDIR /chalk
 COPY *.nimble /chalk/
 
 RUN nimble install --depsOnly --verbose
+
+# Build any static deps not bundled in nimutils' pre-built package directory.
+# Uses the local nimutils tree (mounted as a build context) so that changes to
+# buildlibs.sh (e.g. ensure_zlib) take effect without a nimutils release.
+# COPY --from=nimutils bin/buildlibs.sh /tmp/nimutils-local/buildlibs.sh
+# COPY --from=nimutils files/deps /tmp/nimutils-local/deps
+# RUN bash /tmp/nimutils-local/buildlibs.sh /tmp/nimutils-local/deps
```
viega, line 59: The zlib provisioning is commented out, but `config.nims` now statically links `z`. If the pinned nimutils `files/deps` doesn't bundle `libz.a`, the static build fails at link time. Either uncomment/wire this up, or confirm the deps package ships static zlib and drop the dead lines + a comment explaining where `libz.a` comes from.

src/docker/tar.nim
```diff
@@ -0,0 +1,297 @@
...
+        block:
+          var found = false
+          for component in norm.split('/'):
+            if component.len > 0 and globMatch(component, p):
+              found = true
+              break
+          found
+      else:
+        globMatch(norm, p)
+    if matches:
```
viega: **Fidelity: `.dockerignore` semantics diverge from Docker.**

No-slash patterns are matched against *every path component* (gitignore-style). Docker's `.dockerignore` uses Go `filepath.Match` against the *full relative path*, where `*` never crosses `/` — so `*.log` excludes only top-level `*.log`, and `node_modules` excludes only root `node_modules`, not nested ones. This implementation over-excludes relative to Docker, yet `design-docker-registry.md` and the `honor_dockerignore` field doc both claim it "ensures the uploaded context matches what Docker itself would have used." For the `.dockerignore`-sourced patterns that claim is inaccurate, and over-exclusion silently drops files Docker actually shipped to the build — which undercuts the "exact source files" purpose. Either match Docker's path-anchored semantics for the dockerignore path, or soften the doc claim and document the difference.

src/docker/tar.nim
```diff
@@ -0,0 +1,297 @@
...
+          gz.gzWriteAll(chunk[0 ..< n])
+          written += int64(n)
+        let pad = (tarBlock - int(written mod tarBlock)) mod tarBlock
+        if pad > 0:
+          gzWriteZeros(gz, pad)
+      finally:
+        fh.close()
+
+    else:
+      discard  ## skip symlinks and other special files
```
viega: **Fidelity: symlinks silently dropped.** `else: discard` skips symlinks and special files. Docker build contexts routinely contain symlinks (e.g. `node_modules/.bin`, vendored config links), and `docker build` tars them as symlink entries. Dropping them makes the archive unfaithful and can lose referenced content. Consider emitting symlink entries (ustar typeflag `2`) or at minimum documenting the omission under Limitations.

src/docker/tar.nim
```diff
@@ -0,0 +1,297 @@
...
+    var splitAt = -1
+    let maxSearch = min(name.high - 1, tarNameLen + tarPrefixLen - 1)
+    for i in countdown(maxSearch, 0):
+      if name[i] == '/' and (name.len - i - 1) < tarNameLen:
+        splitAt = i
+        break
+    if splitAt > 0:
+      prefix = name[0 ..< splitAt]
+      name   = name[splitAt + 1 .. ^1]
+    ## If still too long, name is truncated; rare for real contexts.
```
viega: **Silent truncation of long single-component names.** If a path has no `/` available to split into the ustar prefix and the name exceeds 100 bytes, the name is truncated with no error (the comment admits it). Hashed bundle filenames and some `node_modules` artifacts do exceed 100 bytes, producing a wrong path in the tar. Warn, or emit a GNU/PAX long-name record.

src/docker/tar.nim
```diff
@@ -0,0 +1,297 @@
...
+
+## ---------------------------------------------------------------------------
+## Internal helpers
+
+proc toOctal(n: int64, width: int): string =
+  result = newString(width)
+  var v = n
+  for i in countdown(width - 1, 0):
+    result[i] = char(ord('0') + int(v and 7))
+    v = v shr 3
```
viega, line 60: **Silent overflow in `toOctal`.** The 11-digit octal size field caps at 8 GiB; a larger file (possible when `max_file_size` is unset/huge) silently writes a corrupt size header rather than erroring. Add a guard that raises when `n` doesn't fit in `width` octal digits.

src/docker/tar.nim
```diff
@@ -0,0 +1,297 @@
...
+      if isExcluded(norm, patterns):
+        trace("docker: context upload: excluding " & norm)
+        continue
+      let
+        size  = full.getFileSize()
+        mtime = full.getLastModificationTime().toUnix()
+      if maxFileSize > 0 and size > maxFileSize:
+        trace("docker: context upload: skipping large file " & norm &
+              " (" & $size & " bytes > " & $maxFileSize & " bytes)")
+        continue
```
viega, line 587: **Partial archive looks complete.** Files exceeding `max_file_size` are skipped with only a `trace`. The resulting context is partial but nothing surfaces it — no `_OP_FAILED_KEYS` entry, no manifest annotation. For a provenance artifact, a silently-truncated context that *looks* complete is misleading. Consider recording skipped files (count/names) so consumers know the archive is incomplete.

src/docker/tar.nim
```diff
@@ -0,0 +1,297 @@
...
+      full = baseDir / rel
+
+    case kind:
+    of pcDir:
+      if not isExcluded(norm, patterns):
+        let mtime = full.getLastModificationTime().toUnix()
+        gz.gzWriteAll(buildTarHeader(norm, 0, isDir = true, mtime))
+      ## Always recurse so negation patterns can re-include files inside
+      ## an excluded directory (e.g. "logs/" with "!logs/*.log").
+      addDirToTar(gz, baseDir, rel, patterns, maxFileSize)
```
viega: **🟠 Performance: unconditional recursion into excluded dirs.** Always recursing "so negation patterns can re-include files" means that with the default `exclude_patterns: [".git"]` plus typical `.dockerignore` entries (`node_modules`, `.venv`, `dist`), chalk walks, stats, and glob-matches *every file* under those trees just to discard them. On repos with a large `.git` or deep `node_modules` this is a lot of wasted syscalls + pattern matching on the hot path. Prune recursion into an excluded directory unless some negation pattern could plausibly match beneath it (git/Docker both do this). Biggest perf win available here.

src/docker/registry.nim
```diff
@@ -538,13 +539,13 @@ proc request(self:              DockerImage,
...
           if contentType != "":
             headers["Content-Type"] = contentType
-          # only include content-range header not chunked uploads
-          # e.g. ghcr doesnt accept content-range for monolithic uploads
+          # OCI distribution spec requires "{first}-{last}" format (no "bytes " prefix,
+          # no "/{total}" suffix). See:
+          # https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pushing-blobs
+          # Only include for chunked uploads; omit for monolithic uploads because
+          # some registries (e.g. ghcr) reject Content-Range on monolithic PUT.
           if range.a > 0 or range.b > 0 and range.b - range.a + 1 != size:
```
viega: **Fragile operator precedence.** `range.a > 0 or range.b > 0 and range.b - range.a + 1 != size` relies on `and` binding tighter than `or`. It computes the intended result, but please parenthesize for the next reader.

Separately (not this line): `TIMEOUT = 3000` (ms; the `# sec` comment is misleading) is applied per request, including each 5 MB upload chunk in `layerPutFileStream`. On a slow uplink a single chunk can exceed 3s and fail even with `retries = 2`, and contexts can be ~100 MB. Consider scaling the timeout with chunk size or making it configurable now that large bodies flow through this path.

src/docker/context_upload.nim
```diff
@@ -0,0 +1,519 @@
...
+  ## Returns "" if the tarball is within the allowed size.
+  ## Returns a human-readable error message if the size exceeds the threshold.
+  ## The caller is responsible for recording the failure in _OP_FAILED_KEYS.
+  if sizeThreshold == 0:
+    return ""
+  let size = getFileSize(tarPath)
+  if size > sizeThreshold:
+    let msg = (
+      "build context '" & contextName & "' (" & contextPath & ") tarball is " &
+      $size & " bytes which exceeds upload_context_size_threshold of " &
```
viega: **Stale field name in user-facing message.** This references `upload_context_size_threshold`, but the config field is now `size_threshold` under `docker_context_upload` (same stale name in the `addFailedKey` descriptions at lines ~295 and ~504). Users grepping their config for that name won't find it.

src/docker/context_upload.nim
```diff
@@ -0,0 +1,519 @@
...
+        isFetched: true,  # blob already uploaded at build time; skip put()
+      )
+      ctxManifest = newContextManifest(
+        image       = image,
+        subject     = subject,
+        layer       = layer,
+        contextName = contextName,
+      )
+    discard image.appendToAttestationManifestList(ctxManifest)
+    return (ctxManifest.digest.extractDockerHash(), blobSize)
```
viega: **Minor / Limitations:** the `registry` strategy uploads the blob at build time; if the push never lands (build aborts, push disabled, different repo) the blob is orphaned in the registry until GC. Acceptable, but worth a line under Limitations in the design doc.

viega (commented):
## Review: docker build context upload

Well-structured feature with a strong functional-test matrix, standards-correct OCI attachment (`subject` + `sha256-<digest>` referrers fallback tag, fetch-then-append coexistence with sigstore), idempotent HEAD-before-PUT uploads, and a lazy `FileStringStream` that keeps memory flat for large contexts. Also fixes a real latent bug (`build-contents` → `build-context` in `cmdline.nim`, which means `DOCKER_ADDITIONAL_CONTEXTS` never worked before).

My concerns cluster around three areas: **build-system fragility** (static zlib), **archive fidelity vs. what Docker actually built** (the feature's whole purpose is "the exact source files used to produce an image"), and **performance on large real-world contexts**. Inline comments below, roughly in priority order.

### Blocking until verified
- Static `zlib` link added but Dockerfile provisioning commented out — risks the musl/Alpine static build (see `config.nims` / `Dockerfile`).

### Fidelity
- `.dockerignore` matching diverges from Docker's path-anchored `filepath.Match` semantics, yet docs claim it matches Docker.
- Symlinks silently dropped.
- `max_file_size` / oversized-name / >8GB-size truncations happen silently — a partial provenance artifact that looks complete.

### Performance
- Unconditional recursion into excluded dirs (`.git`, `node_modules`) walks + glob-matches every entry just to discard it.
- 3s per-request timeout now governs multi-MB chunk uploads.

Full writeup with rationale is in the inline threads.

viega:
## Build verification (finding #1: static zlib)

I tried to settle the static-zlib concern by running the Linux/Docker static build (`docker compose run --rm chalk nimble release`). Result: **finding #1 is confirmed, and is a hard dependency problem rather than just a doc nit.**

What I found:
- The `chalk` compose service bind-mounts the developer's **local sibling checkouts** of nimutils/con4m (`$PWD/../nimutils`, `$PWD/../con4m`), and this PR adds `additional_contexts: nimutils: ../nimutils`. So the build uses the local nimutils working copy, not the pinned commit.
- In a current nimutils checkout, `bin/buildlibs.sh` has `ensure_*` functions for musl, openssl, pcre, gumbo, hatrack, ffi, and sodium — but **no `ensure_zlib`**, and there is **no `libz.a` anywhere** in the nimutils tree. The only "zlib" reference is OpenSSL's `no-zlib` configure flag.
- Since `config.nims` now adds `"z"` to `staticLinkLibraries(libs, libDir, …)`, the static link needs `libz.a` in `$LOCAL_INSTALL_DIR/libs`, and nothing in the current dep tooling produces it. The Dockerfile hook that *would* run an updated `buildlibs.sh` (`ensure_zlib`) is commented out.

In other words, this PR has an **unpinned, unreleased nimutils dependency**: the static build only links if you happen to have a local nimutils whose `buildlibs.sh` grew an `ensure_zlib` step (or some other provider of `libz.a`). A clean checkout / CI with the currently-pinned con4m→nimutils will fail at link.

**To unblock:** land the `ensure_zlib` buildlibs change in nimutils, bump the con4m/nimutils pin so it's captured, and either wire up (un-comment) the Dockerfile provisioning or confirm the pinned deps package ships `libz.a`. Then the commented-out Dockerfile block can be removed.

(Note: my local build also hit an unrelated `S3SinkState.endpoint` undeclared-field error from a stale local nimutils — that's a local-env skew, not a defect in this PR, since `src/sinks.nim` is unchanged here. It does independently confirm the dep tooling is behind what the branch expects.)

config.nims
miki725: its already bundled in nimutils

src/docker/context_upload.nim
miki725: updated design doc as this is expected behavior

src/docker/registry.nim
miki725: improved

src/docker/tar.nim
miki725: recursing only if possible to negate

src/docker/tar.nim
miki725: added skipped files reporting key to indicate what is missing

src/docker/tar.nim
miki725: raising exception which will bubble up in case this happens

src/docker/tar.nim
miki725: handling long records now

src/docker/tar.nim
miki725: symlinks are honored now

src/docker/tar.nim
miki725: updated to match same synamtics

Dockerfile
miki725: zlib comes from con4m/nimutils

src/docker/tar.nim
```diff
@@ -0,0 +1,625 @@
...
+  result[156] =
+    if typeFlag != '\0':   typeFlag
+    elif isDir:            '5'
+    elif linkTarget != "": '2'
+    else:                  '0'
+
+  ## Linkname field (157..256): symlink target, null-terminated
+  if linkTarget != "":
+    for i in 0 ..< min(linkTarget.len, tarLinkNameLen):
+      result[157 + i] = linkTarget[i]
```
eatonphil, line 164: This will silently truncate long linkTarget names, Claude tells me.

eatonphil: > 1. Symlink target > 100 bytes → silent truncation. The tar writer emits @LongLink preambles for long names (tar.nim:441), but the symlink target is written straight into the 100-byte ustar linkname field with min(linkTarget.len, tarLinkNameLen) (tar.nim:158). A symlink whose target path exceeds 100 chars gets silently corrupted — GNU tar handles this with a typeFlag='K' @LongLink. Test: ln -s "$(python3 -c 'print("a/"*60)')" link in the context, archive, extract, compare target.

src/docker/tar.nim
```diff
@@ -0,0 +1,625 @@
...
+      var entryBytes = int64(tarBlock)
+      if needsLongLink(relPath = norm, isDir = false):
+        entryBytes += gz.writeLongLinkEntry(norm)
+      gz.gzWriteAll(buildTarHeader(
+        relPath = norm,
+        size    = size,
+        isDir   = false,
+        mtime   = mtime,
+      ))
+      let fh = open(full, fmRead)
```
eatonphil, line 597: Claude warns that we don't guard against opening non-regularly files. If it opens a FIFO/named pipe it will just hang here.

> 2. Non-regular files hang the build. walkDir yields FIFOs/device files/sockets as pcFile, and the writer does open(full, fmRead) (tar.nim:539). Opening a FIFO with no writer blocks forever. Test: mkfifo context/pipe; chalk docker build … → expect hang. This is a trivial accidental-DoS (a named pipe in a repo) and an easy deliberate one. Should stat and skip non-regular files.

src/docker/tar.nim
```diff
@@ -0,0 +1,625 @@
...
+      ## Pruning here avoids walking .git, node_modules, etc. unnecessarily.
+      if not excluded or hasNegationForDir(norm, patterns):
+        addDirToTar(gz, baseDir, rel, patterns, maxFileSize, sizeThreshold, skippedFiles, pending)
+
+    of pcFile:
+      if isExcluded(norm, patterns):
+        trace("docker: context upload: excluding " & norm)
+        continue
+      let
+        fss   = newFileStringStream(full)
```
eatonphil: Claude warns of a pathological case of having too many small files.

> 3. fd pressure with many files. Each file is opened twice — once as newFileStringStream(full) for len/sha256 and again via open(full, fmRead) (tar.nim:520,539). If the FileStringStream fd isn't promptly closed, a 100k-file context can exhaust the fd table. Test: context with 200k tiny files; watch lsof/EMFILE.

src/docker/tar.nim
```diff
@@ -0,0 +1,625 @@
...
+        size  = int64(len(fss))
+        mtime = full.getLastModificationTime().toUnix()
+      if maxFileSize > 0 and size > maxFileSize:
+        let hash = fss.sha256Hex()
```
eatonphil: Claude tells me this sha256Hex() calculation reads the whole file even when we're going to skip it below.

src/docker/tar.nim
miki725: file streaming api will handle that correctly out of the box. cant use the file string stream for the tar stuff but each file is closed after its written so should be ok for large contexts

src/docker/tar.nim
```diff
@@ -0,0 +1,625 @@
...
+        return true
+  return false
+
+proc checkSizeThreshold(gz: GzFile, sizeThreshold: int64) =
+  ## Unconditionally flush gzip output and raise TarSizeLimitError if the
+  ## compressed output exceeds sizeThreshold.  Used at end of archive.
+  if sizeThreshold <= 0:
+    return
+  discard gzflush(gz, 2)  ## Z_SYNC_FLUSH: flush to fd so gzoffset is accurate
+  if gzoffset(gz) > sizeThreshold:
```
eatonphil, line 461: Claude warns about compression bombs not being checked here

> 12. Compression bomb. size_threshold checks compressed gzoffset (tar.nim:417), so 50 GB of zeros sails under the default 100 MB threshold and uploads a tiny tarball that detonates on whatever extracts it downstream (Ocular). Worth deciding whether an uncompressed bound is also needed.

src/docker/tar.nim
eatonphil: > 11. Pathological filenames through the tar writer: names with embedded newlines, backslashes, a file literally named ././@LongLink, a component exactly 100/101/155/255/256 bytes, a single 200-char path component with no / to split on (the code comment admits "name is truncated" — tar.nim:112).

A few cases Claude called out to consider.

src/docker/tar.nim
miki725: yes thats correct behavior by design. if we skip the file we report the skipped files hash so at least we have a concrete reference of what exactly was skipped. otherwise the attestation of the context is incomplete and ambiguous

src/docker/manifest.nim
```diff
@@ -717,6 +717,28 @@ proc put*(self: DockerManifest) =
...
 
+proc appendToAttestationManifestList*(image: DockerImage,
+                                      item: DockerManifest,
+                                      ): DockerManifest {.discardable.} =
+  ## Fetch (or create) the OCI attestation manifest list for `image` and append
+  ## `item` to it, then push the updated list.  `image` must have a digest set.
+  ## Returns the manifest list that was pushed.
+  let attSpec = image.asOciAttestation()
+  try:
+    result = attSpec.fetchListManifest(fetchManifests = true)
```
eatonphil, line 728: Claude highlights potential race here

> 7. Read-modify-write race on the attestation list. appendToAttestationManifestList does fetch→add→put with no concurrency control (manifest.nim:720). Two platforms (or two machines) completing context uploads for the same attestation tag → lost update. Also check idempotency: re-running a completed build/push — does result.add(item) dedup, or accumulate duplicate layers in the index?

src/docker/tar.nim
miki725: chalk just cares about the tar size of what will needed to be uploaded to the registry. if the context is too large we dont want to be uploading multi-gb contexts...

if needed we can add a config for the uncompressed size too but dont think its needed here. ocular is meant to scan the build context as-is

src/docker/registry.nim
```diff
@@ -770,14 +789,15 @@ proc layerPutFileStream*(layer:       DockerImage,
...
-      httpMethod = HttpPatch
-      response:    Response
-    while startAt < size - 1:
+      (location, minChunkSize) = layer.layerPutStart()
+      effectiveChunkSize       = max(chunkSize, minChunkSize)
+      startAt                  = 0
+      attempts                 = 1
+      httpMethod               = HttpPatch
+      response:                  Response
+    while startAt < size:
```
eatonphil, line 830: > 6. Zero-byte blob. while startAt < size never executes for a 0-byte body, leaving response nil → crash on a registry that doesn't already have the empty blob. Can't trigger via context tars, but layerPutString/layerPutJson with empty body can.

src/docker/registry.nim
eatonphil: Claude asks about handling mailicious responses

> 5. Hostile registry responses (registry.nim): omit Docker-Content-Digest in the final response (→ nil-deref crash at the return); return oci-chunk-min-length: -5, : garbage, : 99999999999; return a Range header that decrements or repeats (attempts cap should catch — verify it actually does and doesn't livelock); return 416 with no Location; return a different digest than what was uploaded (does anything detect substitution, or is the registry trusted?). A mock HTTP server returning adversarial headers is the right harness here.

tests/equivalency/dockerignore/testcases.json
```diff
@@ -0,0 +1,614 @@
+[
+  {
```
eatonphil, line 2: Claude suggests checking malformed dockerignores too since these might come up in development

> > 10. Dockerignore/glob malformed patterns: [ unterminated, [], [!], [z-a] (reversed range), [a-], trailing \, patterns that are just / or !, ! alone, deeply nested **. The moby equivalency harness already exists — I'd add these to testcases.json and diff Nim vs Go rather than eyeballing.

src/docker/registry.nim
miki725: technically an edge case but cant happen in chalk as all upload cases in chalk are non-zero. will add a guard though

src/docker/build.nim
```diff
@@ -614,6 +614,28 @@ proc dockerBuild*(ctx: DockerInvocation): int =
   ctx.processPlatforms()
   ctx.pinBuildSectionBaseImages()
 
+  cleanBuildContextCache()
```
eatonphil, line 617: 1. TTL GCs an in-flight / not-yet-pushed tar (local strategy)
The cache is bucketed by a second-granularity wall-clock dir name, and the GC deletes whole buckets older than the TTL:

Dir name format — "yyyy-MM-dd'T'HH-mm-ss" (context_upload.nim:36), stamped at tar-write time: dateDir = contextCacheDir() / now().utc.format(...) (context_upload.nim:175).
The GC itself — cleanBuildContextCache parses each subdir's name and removeDirs it when dt.toTime() < cutoff, where cutoff = getTime() - build_context_cache_max_age (context_upload.nim:80–101, TTL read at :86).
Called at the start of every build (build.nim:617) and every push (push.nim:80).
The hole: the tar must survive from build until a separate chalk docker push completes (context_upload.nim:435–470 reads it back). If build_context_cache_max_age is shorter than the build→push gap — or another build/push fires cleanBuildContextCache in between — the bucket holding the still-needed tar is deleted. The code even anticipates this with the error string at context_upload.nim:441 ("may have been removed by the build_context_cache_max_age TTL"), so it's a known failure surface, just unbounded by anything other than operator-set TTL.

src/docker/context_upload.nim
```diff
@@ -0,0 +1,626 @@
...
+      raise newException(
+        ValueError,
+        "context tarball not found at push time: '" & tarPath & "'; " &
+        "it may have been removed by the build_context_cache_max_age TTL " &
+        "if too much time elapsed between build and push",
+      )
+    let
+      storedHash = snapshot{"tar_hash"}.getStr("")
+      actualHash = newFileStringStream(tarPath).sha256Hex()
+    if storedHash != "" and actualHash != storedHash:
```
eatonphil, line 447: 2. Same-second, same-CHALK_ID filename collision
createDir(dateDir) is idempotent — two invocations in the same second share the bucket (context_upload.nim:176).
The filename is fully deterministic: chalkId & "-" & configSlug & "-" & contextSlug & "-" & contextHash & ".tar.gz" (context_upload.nim:181) — no PID, no nonce, no counter.
So two concurrent chalk docker builds of the same artifact (same CHALK_ID, same registry/push config, same context name) within one wall-clock second resolve to the identical outPath, and writeTarGz (context_upload.nim:183) opens it "wb" — the two writers interleave/clobber. For the local strategy that's worse than transient: the recorded tar_hash (context_upload.nim:331) is computed right after writing, so build A can stamp a hash for a file build B then overwrites, and push time fails the hash-mismatch check (context_upload.nim:447–453) — or silently uploads B's bytes under A's snapshot.

src/docker/context_upload.nim
```diff
@@ -0,0 +1,626 @@
...
+          "strategy":      "local",
+          "tar_path":      tarPath,
+          "tar_hash":      newFileStringStream(tarPath).sha256Hex(),
+          "skipped_files": skippedFilesToJson(skippedFiles),
+        }
+        trace("docker: build context tarball cached at: " & tarPath)
+
+      of "disk":
+        # Record the context path; push time will read from disk.
+        entry = %*{
```
eatonphil, line 338: 3. disk strategy: context mutated/deleted between build and push
The disk strategy defers all tarring to push time and re-reads the live directory:

Existence guard only — raises if the dir is gone (context_upload.nim:472–480).
A bare warn that contents may have drifted, then it re-tars from disk anyway (context_upload.nim:481–502).
Unlike local, there's no integrity anchor — no stored hash to compare against (local has tar_hash; disk has nothing). So rm -rf → clean error; but echo pwned >> context/file between build and push → silently attests a tarball that no longer matches what was built. For an attestation feature that's the more dangerous of the two: the snapshot fields for disk (context_upload.nim:338–348) carry config but no content digest to detect the drift.

miki725: this is by design. disk mode is not the default but is meant for users who do not wish to use any resources during build but simply the dir as-is during upload. for example in CI like:

```
docker build ...
docker push ...
```

if nothing mutates the context, dist strategy might be valid here. however due to the integrity limitation its not a default and is warned that drift can occur in the config docs

src/docker/context_upload.nim
miki725: chalk id is random in docker. this is impossible

src/docker/manifest.nim
miki725: registry has no notion of atomic appends so this is pretty much the best we can do although Im currently adding support for referrers API in which case the tag wont be as relevant anymore as all attestations will be queried via the subject api instead

src/docker/build.nim
miki725: valid edge case but we need to cleanup at some point otherwise we can accumulate a bunch of tmp files so nothing we can do

tests/equivalency/dockerignore/testcases.json
miki725: addressed in [`26ccf63` (this PR)](https://github.com/crashappsec/chalk/pull/669/commits/26ccf635ee767cac0503e10ce71b91400d0f536f)

src/docker/registry.nim
miki725: hardened in https://github.com/crashappsec/chalk/pull/669/commits/26ccf635ee767cac0503e10ce71b91400d0f536f

src/docker/registry.nim
miki725: added https://github.com/crashappsec/chalk/pull/669/commits/26ccf635ee767cac0503e10ce71b91400d0f536f

src/docker/tar.nim
miki725: added https://github.com/crashappsec/chalk/pull/669/commits/26ccf635ee767cac0503e10ce71b91400d0f536f

src/docker/tar.nim
miki725: added https://github.com/crashappsec/chalk/pull/669/commits/26ccf635ee767cac0503e10ce71b91400d0f536f

eatonphil (approved):

PR #673, opened 2026-06-09
fix: switch to use only object store and don't base64 encode data

nettrino:
with the new object store capabilities, we don't need to base64 observables data at all

### DIFF
--- a/configs/co/build_observables_key.c4m
+++ b/configs/co/build_observables_key.c4m
@@ -1,7 +1,8 @@
 keyspec _X_BUILD_OBSERVABLES {
-  kind:  RunTimeHost
-  type:  string
-  callback: func build_observables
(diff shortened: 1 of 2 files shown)


vatlidak (approved):

miki725 (changes requested):
This will break existing chalk versions unless we want to only allow bobs for latest chalk version in chalkapi

nettrino:
@miki725 is this because the configs will be getting applied to older versions too? can we pin the config change from a version up?

nettrino:
I think its fine to only allow bobs in the latest chalk version btw if that is easier to do - I am not sure how to pin this on the config level though

miki725 (approved):
updated the config to be compatible with older versions

PR #674, opened 2026-06-16
feat: OCI referrers API support for attestations

miki725:

## Issue

Immutable tags registry fails when both context upload and attestations are enabled

## Description

When a registry supports the OCI Referrers API (`GET /v2/{name}/referrers/{digest}`),
attestations are now pushed as referrer manifests using the `subject` field rather than
relying solely on the legacy `sha256-<digest>` tag schema. Support is detected
automatically via the `OCI-Subject` response header on manifest operations.

Changes:

- **`registry.nim`**: detect referrers support via `OCI-Subject` response header and
  cache per repo; add `referrersGet` (fetch referrers index) and `supportsReferrers`
  (cached probe) procs
- **`manifest.nim`**: replace `appendToAttestationManifestList` with `addAttestation`
  which pushes the referrer manifest and optionally also maintains the legacy tag;
  `addAttestationToTag` now defers the tag push into `pendingAttestationTags`;
  `flushAttestationTags` pushes all pending tags at the end of build/push to avoid
  duplicate writes to immutable-tag registries (e.g. when both context upload and
  sigstore attestation run in the same operation)
- **`build.nim`**, **`push.nim`**: call `flushAttestationTags` after all attestations
  have been added
- **`context_upload.nim`**, **`attestation_api.nim`**: updated to use `addAttestation`
- **`chalk.c42spec`**: new `docker.attestation_use_oci_tag` field — when the referrers
  API is available, also maintain the legacy tag for clients that only understand the
  tag schema (default: `true`); set to `false` to skip the tag push entirely

## Testing

```
maketests test_docker.py
```

### DIFF
--- a/src/docker/build.nim
+++ b/src/docker/build.nim
@@ -34,6 +34,7 @@ import "."/[
   image,
   inspect,
   login,
+  manifest,
   platform,
   registry,
   scan,
@@ -781,6 +782,7 @@ proc dockerBuild*(ctx: DockerInvocation): int =
         chalk.collectedData["_OP_ARTIFACT_CONTEXT"] = pack("build")
         chalk.collectRunTimeArtifactInfo()
   collectRunTimeHostInfo()
+  flushAttestationTags()
 
   if wrapVirtual and result == 0:
     for platform, chalk in chalksByPlatform:
--- a/src/docker/manifest.nim
+++ b/src/docker/manifest.nim
@@ -33,8 +33,9 @@ type FilterManifests = ref object
 
 # cache is by repo ref as its normalized in buildx imagetools command
 var
-  jsonCache     = initTable[string, DigestedJson]()
-  manifestCache = initTable[string, DockerManifest]()
+  jsonCache              = initTable[string, DigestedJson]()
+  manifestCache          = initTable[string, DockerManifest]()
+  pendingAttestationTags = initOrderedTable[string, DockerManifest]()
 
 proc getCompressedSize(self: DockerManifest): int =
   if self.kind != DockerManifestType.image:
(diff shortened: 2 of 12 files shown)


src/docker/build.nim
```diff
@@ -781,6 +782,7 @@ proc dockerBuild*(ctx: DockerInvocation): int =
         chalk.collectedData["_OP_ARTIFACT_CONTEXT"] = pack("build")
         chalk.collectRunTimeArtifactInfo()
   collectRunTimeHostInfo()
+  flushAttestationTags()
```
eatonphil, line 785: 1. flushAttestationTags() is not crash-safe — attestations can be silently lost on non-referrers registries. (build.nim:785, push.nim:90)

The referrer put() writes the attestation manifest by digest only — it's unreferenced until either the registry's referrers index picks it up (via subject) or the legacy tag is written at flush. On a registry without referrers support, the only thing that makes that manifest discoverable is the deferred tag push. But flushAttestationTags() is a plain statement, not in a finally, and dockerBuild has an early return at build.nim:763 on push failure. If anything between the attestation push and the flush throws, you've written a dangling manifest that the registry will eventually GC — attestation lost, no error surfaced. This is the finding I'd most want addressed: the flush belongs in a finally.

src/docker/manifest.nim
eatonphil, line 933: 2. Backward-compat read gap in fetchAttestationManifests. (manifest.nim:923)

When supportsReferrers(subject) is true and referrersGet returns a non-nil (but possibly empty) index, it returns immediately and never falls back to the legacy sha256-<digest> tag. Attestations written by an older chalk as tag-only (no subject field, so not in the referrers index) become invisible when verified against a registry that has since gained referrers support. The fallback only triggers when referrersGet returns nil, not when it returns an index that's missing the attestation.

src/docker/registry.nim
```diff
@@ -638,7 +646,8 @@ proc manifestHead*(image:    DockerImage,
       ),
     )
     contentType = response.headers["Content-Type"]
-    digest      = validateDigest(response.headers["Docker-Content-Digest"])
+    digest      = response.headers["Docker-Content-Digest"]
```
eatonphil: 5. validateDigest was dropped from manifestHead. (registry.nim:649) digest = validateDigest(...) became digest = response.headers["Docker-Content-Digest"] — the registry-supplied digest is now trusted unvalidated. Was this intentional (some referrer/attestation response that failed validation), or collateral?

src/docker/manifest.nim
```diff
@@ -252,7 +253,10 @@ proc newManifest(name: DockerImage, data: DigestedJson): DockerManifest =
     )
     list.setJson(data)
     list.setAnnotations(json)
-    for item in json["manifests"].items():
+    let manifests = json{"manifests"}
+    if manifests == nil or manifests.kind != JArray:
+      return list
```
eatonphil: 6. newManifest now silently returns an empty list for missing/non-array manifests. (manifest.nim:256) Good defensiveness against the old KeyError, but it also masks a malformed referrers-index response as "zero attestations" rather than an error — interacts with finding #2.

src/docker/build.nim
miki725: this is expected behavior. even currently if the tag creation fails, chalk just logs it as an error but then keeps going. otherwise it would break the build which is the most important principle in chalk - never fail a build

src/docker/manifest.nim
miki725: oh this is a valid point. will fix

src/docker/registry.nim
miki725: rebase issue. updating

src/docker/manifest.nim
miki725: improved in [`7b5d3cf` (this PR)](https://github.com/crashappsec/chalk/pull/674/changes/7b5d3cfe015ce4892e24c9b56d0f5d6a44155856)

src/docker/registry.nim
miki725: fixed in [`7b5d3cf` (this PR)](https://github.com/crashappsec/chalk/pull/674/changes/7b5d3cfe015ce4892e24c9b56d0f5d6a44155856)

src/docker/manifest.nim
miki725: improved in [`7b5d3cf` (this PR)](https://github.com/crashappsec/chalk/pull/674/changes/7b5d3cfe015ce4892e24c9b56d0f5d6a44155856)

eatonphil: 1. Latent correctness: artifactType read from the wrong JSON level
manifest.nim:273, inside newManifest's list branch:

artifactType: json{"artifactType"}.getStr(),   # top-level index
This reads the index's top-level artifactType for every child descriptor instead of item{"artifactType"}. In an OCI referrers index each descriptor carries its own artifactType; the top level usually has none. This was dormant before (normal manifest lists don't set it), but fetchAttestationManifests now feeds a referrers index through newManifest, so it's newly exercised. The current read path still works (filtering is by annotations, and referrersGet is always called with artifactType = ""), so this is latent rather than breaking, but it's wrong and will bite anyone who later filters referrers by artifact type. Fix: item{"artifactType"}.getStr().

Related: the server-side filter path in referrersGet (the ?artifactType= query branch, registry.nim:967) is never called with a non-empty value anywhere, so that URL-building (concatenating ?artifactType= onto path before combine(defaultUri, url)) is untested. Low risk while unused, but dead-ish code.

src/docker/manifest.nim
```diff
@@ -716,28 +727,65 @@ proc put*(self: DockerManifest) =
...
+
+  item.name = subject.withBare()
+  item.link()
+  info("attestation: pushing attestation for " & $subject)
+  item.put()
+
+  # Check after put() so that the OCI-Subject response header from the
+  # registry has a chance to warm the referrers cache before we decide
+  # whether to also push the legacy sha256-<digest> tag.
+  if not supportsReferrers(subject) or useOciTag:
```
eatonphil: 3. Wasted referrers probe in the default config
manifest.nim:784:

if not supportsReferrers(subject) or useOciTag:
useOciTag defaults to true, in which case the tag is pushed regardless of referrers support, yet supportsReferrers(subject) is still evaluated first. On a registry that does not support referrers (so the OCI-Subject header never warmed the cache), this forces an extra referrersGet HTTP probe whose result is then ignored. Reorder to short-circuit:

if useOciTag or not supportsReferrers(subject):
Correctness is unchanged; it just skips a pointless round-trip per repo in the default config.

src/docker/registry.nim
```diff
@@ -639,6 +647,7 @@ proc manifestHead*(image:    DockerImage,
     )
     contentType = response.headers["Content-Type"]
     digest      = validateDigest(response.headers["Docker-Content-Digest"])
+  checkReferrersSupport(image, response)
```
eatonphil, line 651: 4. checkReferrersSupport in HEAD/GET is dead
The new checkReferrersSupport calls in registry.nim:650 (manifestHead) and registry.nim:683 (manifestGet) can never fire: OCI-Subject is a push-response header, so HEAD/GET responses don't carry it. Only the manifestPut call site (registry.nim:955) is meaningful. Harmless, but misleading: I'd drop the two read-path calls or add a comment, otherwise a future reader assumes reads warm the cache.

src/docker/manifest.nim
miki725: fixed in [`ce7332b` (this PR)](https://github.com/crashappsec/chalk/pull/674/changes/ce7332b2dec67309dfeeb4c609fee54f26446143)

src/docker/manifest.nim
miki725: fixed in [`adbf5c1` (this PR)](https://github.com/crashappsec/chalk/pull/674/changes/adbf5c1076da319d61d29fa4404c6860b55e06e3#diff-c61c95db1ae78e1ba408d27f1d3525e86521bcd4c977e4c82299085d44d41d4f)

src/docker/registry.nim
miki725: added comment in [`adbf5c1` (this PR)](https://github.com/crashappsec/chalk/pull/674/changes/adbf5c1076da319d61d29fa4404c6860b55e06e3#diff-c61c95db1ae78e1ba408d27f1d3525e86521bcd4c977e4c82299085d44d41d4f)

we should still check if it will send the header by any chance

eatonphil (approved):

PR #675, opened 2026-06-16
update the bundle path according to curiosity updates

vatlidak:


miki725 (commented):
not all of the commands will be guaranteed to exist on the runners. it prolly will on the github hosted ones but self-hosted or container based runners some of these might not exist

otherwise looks good

miki725 (approved):

PR #682, opened 2026-06-24
fix: handle ECR stale Range header in chunked blob uploads

miki725:
## Summary

- ECR accepts each PATCH chunk correctly but reports a non-advancing `Range` response header, causing spec-compliant clients to re-send data at the wrong offset and receive a `416 Range Not Satisfiable`. After one failed retry, Chalk detects the stall and switches to tracking upload position by what was sent rather than what the registry reports.
- Adds `httpMethod` reset per iteration so a failed final PUT does not leave subsequent intermediate chunks as `HttpPut`.
- Adds an attempt limit on the `416` retry path to prevent an infinite loop when a chunk is genuinely rejected.
- Adds `test_build_context_upload_ecr` — a functional test that uploads 33 MB of incompressible data to ECR, exercises the multi-chunk stall-detection path end-to-end, and verifies content byte-for-byte via SHA-256.

References the upstream ECR bug: [aws/containers-roadmap#2831](https://github.com/aws/containers-roadmap/issues/2831)

## Test plan

```shell
# ECR functional test (requires AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_ACCOUNT_ID)
make tests args="test_docker.py::test_build_context_upload_ecr -x -s"

# Existing context upload tests
make tests args="test_docker.py::test_build_context_upload -x"

# Unit tests
make unit-tests args="pattern tests/unit/test_registry_put_file_stream.nim"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 # Chalk Release Notes
 
+## 1.1.2
+
+### Bug Fixes
+
+- Build context upload to ECR no longer fails for blobs larger than one chunk.
+  ECR stores each PATCH chunk correctly but reports a stale `Range` header
+  that does not advance, causing spec-compliant clients to re-send chunks at
+  the wrong offset and receive a `416`. Chalk now detects a non-advancing
+  `Range` after one retry and switches to tracking position by what was sent
+  rather than what the registry reports.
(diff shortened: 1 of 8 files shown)


baron-crashoverride (approved):
lgtm

PR #684, opened 2026-07-06
fix: guard macho codec against insufficient LC header slack

miki725:
## Summary

- Binaries compiled without `-headerpad,0x1000` can have as little as 24–39 bytes of LC header padding. The native Mach-O codec was silently corrupting these: `strip_signature()` frees 16 bytes, chalk's `LC_NOTE` costs 40 bytes, then `codesign` needs 16 bytes back for `LC_CODE_SIGNATURE` — leaving it no room and producing a binary that fails to run on macOS.
- For `csNone` binaries with < 40 bytes of slack, the C-side check correctly refused but `machoScan` had already claimed the artifact, so the script-wrapper codec never got to try.
- Adds `chalk_macho_lc_slack()` (C + Nim FFI) and checks it in `machoScan` before claiming a fresh-chalk artifact: if pre-strip slack < 40 bytes, returns `none(ChalkObj)` so the wrapper codec handles the binary instead.

## Test plan

```shell
# Build a binary without -headerpad and confirm chalk falls back to the
# wrapper codec rather than producing a non-runnable binary.
# On a macOS host with codesign available:
clang -Wl,-headerpad,0x10 -o /tmp/tight_binary /path/to/source.c
chalk insert /tmp/tight_binary   # should wrap, not corrupt
/tmp/tight_binary                # should still run
chalk extract /tmp/tight_binary  # should find the chalk mark
```


eatonphil:
Works for me! Two notes from claude:

> No test for the new path. test_macho.nim exercises the happy path on a fixture with ample headerpad, but nothing covers lcSlack() or the < 40 B → defer behavior — the exact regression this PR fixes. A fixture compiled without -headerpad (or a synthetic one with trimmed padding) asserting parsed.lcSlack() < 40 would lock in the fix. Worth adding before merge, or as a fast-follow.

> lcSlack() recomputes the layout C-side (a second compute_layout walk), duplicating work add_note does again later. Negligible cost, not worth changing — just noting it's not free.

eatonphil (approved):
tested and works

PR #762, opened 2026-08-10
feat: add Dockerode post-push telemetry

### LINKED ISSUE #761: Support post-push telemetry for Dockerode image pushes
### Requirements

The authoritative requirements moved to [Dockerode post-push requirements](https://github.com/crashappsec/erics-status-and-requirements/blob/ae49fab079c92f9718ec2214fc09d0f38acb2222/docs/chalk/dockerode-post-push.md) at central requirements commit `ae49fab079c92f9718ec2214fc09d0f38acb2222`.

This issue continues to track the product problem. Implementation is in [#762](https://github.com/crashappsec/chalk/pull/762).

DavidsonErickson:
Closes #761

## Summary

This update adds fail-open post-push telemetry for one Dockerode image push. It also adds an internal installer for central activation on one managed Linux Jenkins agent class.

The consumer repository stays unchanged. Its Jenkinsfile still runs the exact `npm run publish` command without a Chalk prefix or a new dependency.

The agent bootstrap runs `chalk __ dockerode_policy_install --root <absolute-path>` before the inbound agent connects. The installer copies Chalk and the embedded loader files into a release directory named by the Chalk binary SHA-256. It verifies file hashes and modes, writes a pinned activation script, preserves existing `NODE_OPTIONS`, and rejects relative roots, symlink paths, incomplete releases, corrupt files, and unsafe permissions.

`chalk dockerode -- <command>` remains available for manual activation. Manual and central activation use the same embedded loader bytes. If both activate in one process tree, the global patch marker prevents a second patch, Engine push, or Chalk report.

Dockerode performs the Engine push once. Chalk waits only for post-push work. Loader, parser, spawn, handoff, and timeout failures do not replace Dockerode's return value or error. Chalk never retries or replays the Engine push.

## Version boundary

The customer confirmed runtime `dockerode@3.3.5` on September 4, 2026. This candidate pins that public runtime and the Docker Modem 3.x boundary. The tested dependency tree contained `docker-modem@3.0.8`. `@types/dockerode@3.3.47` supplies compile-time declarations only; the runtime is no longer unidentified. Earlier Dockerode 5.0.1 evidence is historical and does not establish 3.3.5 support. Other Dockerode runtime versions remain unsupported.

The reviewed Node majors are 20, 22, 24, and 26. Node 22 requires 22.15.0 or newer. Node 20 uses a bounded CommonJS loader path. Node 22.15.0 and newer reviewed majors use synchronous `node:module.registerHooks`. Other Node releases stay disabled. The repository's exact configured version requires validation before trial use.

The remaining runtime boundary is Linux, Dockerode's CommonJS `lib/image.js`, the foreground npm process tree, `/var/run/docker.sock`, one explicit tag, and Dockerode's username, password, and server-address auth shape. Bootstrap validation must pass before the inbound agent connects.

## Validation

[GitLab pipeline 2821443427](https://gitlab.com/crashappsec/secureworks-poc-repos/chalk-docker-reference/-/pipelines/2821443427) passed on September 4, 2026. It built Chalk commit `a461c8f978cffb0934c033b5040fc81c66c858dd`, which matches this PR's head checked on September 14, and used consumer commit `b09b582bf441f7ca37ab847bf0e1d5534a000243`.

- Node 20, 22, and 24 passed the central-loader compatibility jobs. [Node 26 job 16316888688](https://gitlab.com/crashappsec/secureworks-poc-repos/chalk-docker-reference/-/jobs/16316888688) passed the full temporary Jenkins proof owned by Crash Override, without Cvent access.
- Every arm ran the exact `npm run publish` command. Consumer file hashes matched before and after each run, and the controller's embedded Jenkinsfile matched the repository's SHA-256.
- Baseline, bootstrap activation, one-millisecond timeout, and rollback each pushed once. Their Chalk push-report counts were 0, 1, 0, and 0. Activation emitted `posthook_complete`; timeout emitted `posthook_timeout` and preserved caller success and registry presence.
- Bootstrap verified the immutable policy release before each activated agent connected. Rollback restarted the agent without policy; reinstallation resolved to the same immutable activation.
- Dockerode, the GitLab registry, local `RepoDigests`, and the Chalk push report agreed on `sha256:a57a26450eb3e3f2456e0933a60e79b10c21c178459d670ac3e9d0eae390e967`.
- Image execution and the exact-secret scan passed. Retained archive checksums verified.
- The retained independent Frank Guppy export confirms build, push, exec, and a matching heartbeat. The push row binds the tested commits, registry digest, job, and `npm run publish` ancestry. Heartbeat reporting was explicitly enabled. This run established no Chalk pull report.

The [current-status record](https://github.com/crashappsec/erics-status-and-requirements/blob/8dfe6a6/docs/chalk/dockerode-current-status-2026-09-04.md) and [retained Frank Guppy export](https://github.com/crashappsec/erics-status-and-requirements/blob/8dfe6a6/docs/chalk/dockerode-335-frank-guppy-evidence-2026-09-04.md) contain the evidence and exact queries.

## Scope

This PR does not add generic Jenkins policy, controller plugins, support for another SDK, support for other Dockerode runtime versions, multi-agent rollout, a production timeout, or Cvent production deployment.

The GitLab Jenkins proof and independent Frank Guppy push receipt are complete. The central-policy ECR and Docker Desktop acceptance path, Chalk pull telemetry, intended Cvent agent and tenant, production attestation delivery, broader rollout, merge, and release remain outside this proof. Any Cvent infrastructure use requires a separately agreed trial. The candidate remains open, unmerged, and unreleased.

The requirements file is [Dockerode post-push requirements](https://github.com/crashappsec/erics-status-and-requirements/blob/028a19f7c03b8e8bd7135ff3fd01c139fa5c1092/docs/chalk/dockerode-post-push.md).


src/commands/cmd_internal.nim
```diff
@@ -6,18 +6,132 @@
...
+  if not attrGet[bool]("load_external_config"):
+    putEnv("CHALK_DOCKERODE_NO_EXTERNAL_CONFIG", "1")
+
+  try:
+    quitChalk(runCmdNoOutputCapture(args[0], commandArgs))
+  finally:
+    try:
+      removeDir(loaderDir)
+    except:
+      warn("dockerode_run: could not remove temporary loader directory")
```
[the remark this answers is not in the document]
DavidsonErickson: Fixed in fd1e7ca. runDockerodeCommand now returns the child exit code only after its finally block removes the private loader directory; quitChalk is called afterward. The new native regression test exercises both exit 0 and exit 7 and asserts that no loader directory remains.

src/docker/push.nim
```diff
@@ -11,18 +11,71 @@ import ".."/[
...
+    local        = inspectImageJson(image)
+    expectedRepo = parseImage(image).repo
+  var digestMatches = false
+  for repoDigest in local{"RepoDigests"}.getStrElems():
+    if repoDigest == expectedRepo & "@" & expectedDigest:
+      digestMatches = true
+      break
+  if not digestMatches:
+    error("docker: post-push digest does not match local RepoDigests for " & image)
+    return false
```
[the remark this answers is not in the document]
DavidsonErickson: Fixed the confirmed normalization issue in fd1e7ca. RepoDigest matching now parses both sides and compares normalized registry, repository name, and digest, with coverage for Docker Hub aliases, a custom registry/port, empty RepoDigests, and mismatches. We also verified that getStrElems(nil) is already safe, so the empty-list regression captures that existing behavior.

DavidsonErickson:
GitHub UI Copilot PR-lite review disposition:

- Fixed process-environment leakage in signed `ceaabb6`; the wrapper now restores `NODE_OPTIONS` and both `CHALK_DOCKERODE_*` variables, with native regression coverage.
- Fixed empty auth-field handling in `ceaabb6`; unsupported auth now skips instrumentation before invoking the post-hook, with Node contract coverage.
- Bounded the internal stdin handoff to 64 KiB in signed `0a53a6a`; oversized input returns `unsupported_input` / exit 2, with native and compiled-binary proofs.
- Fixed malformed or missing-field input classification in signed `309ab3b`; parse/type failures now return the documented `unsupported_input` / exit 2, while post-processing exceptions remain exit 1. Direct compiled-binary proofs cover malformed JSON and missing fields.
- No change for the suggestion to await `nextLoad` in `register.cjs`: this code uses `module.registerHooks()`, whose hooks and `nextLoad` are explicitly synchronous. The Promise-returning API belongs to `module.register()`. Making this hook async would violate the selected Node API and break synchronous CommonJS `require()` coverage.

Validation: 16/16 Node contracts, all 20 native unit programs, release build, diff checks, and signed-commit verification passed locally. Fresh hosted checks are running on the final head.

src/dockerode_post_push/runner.nim
```diff
@@ -0,0 +1,66 @@
...
+    hadNoExternal      = existsEnv("CHALK_DOCKERODE_NO_EXTERNAL_CONFIG")
+
+  try:
+    discard chmod(cstring(loaderDir), Mode(0o700))
+    writeFile(register, dockerodeRegisterSource)
+    writeFile(runtime, dockerodeRuntimeSource)
+    discard chmod(cstring(register), Mode(0o600))
+    discard chmod(cstring(runtime), Mode(0o600))
+    putEnv("NODE_OPTIONS", (priorNodeOptions & " --require=\"" & register & "\"").strip())
+    putEnv("CHALK_DOCKERODE_CHALK", getMyAppPath())
```
DavidsonErickson: **Wrapper never sets the deadline env var it requires.**

`runDockerodeCommand` sets `NODE_OPTIONS`, `CHALK_DOCKERODE_CHALK` and `CHALK_DOCKERODE_NO_EXTERNAL_CONFIG`, but not `CHALK_DOCKERODE_POST_PUSH_TIMEOUT_MS`. `runtime.cjs:181-182` returns `{supported:false, code:'unsupported_no_deadline'}` when that variable is undefined, so `supportedOperation` bails before the socket check: no payload is written, `chalk __ docker_post_push` is never spawned, and no push report is emitted.

As shipped this makes the feature a silent no-op by default. Either set a default here, or make the deadline optional in `postPushDeadline()` with a documented fallback.

src/dockerode_post_push/register.cjs
```diff
@@ -0,0 +1,71 @@
...
+    } catch (_) {
+      // Most ancestor directories are not package roots.
+    }
+    dir = path.dirname(dir);
+  }
+  return null;
+}
+
+const [nodeMajor, nodeMinor] = process.versions.node.split('.').map(Number);
+const supportedNode = (nodeMajor === 22 && nodeMinor >= 15) || nodeMajor === 24;
```
DavidsonErickson: **Exact-major allowlist silently disables instrumentation on newer Node.**

`(nodeMajor === 22 && nodeMinor >= 15) || nodeMajor === 24` excludes Node 23, 25, 26 and every future release. `registerHooks` has been stable since 22.15, so this is a version-gate bug rather than a capability gate: on an unlisted major, `registerHooks` is never called, `lib/image.js` is never patched, and every push loses telemetry with no diagnostic record and no non-zero exit.

A minimum-version check (`nodeMajor > 22 || (nodeMajor === 22 && nodeMinor >= 15)`) plus the existing `typeof registerHooks !== 'function'` guard would fail closed on capability instead of on version.

DavidsonErickson (commented):
## Review: correctness pass plus end-to-end artifact verification

Reviewed the full diff at high effort (independent finder passes per correctness angle, each candidate re-verified by a separate pass before reporting). Ten correctness findings below, eight confirmed and two plausible. Twelve lower-severity cleanup items were dropped under a reporting cap.

### The headline

Three findings mean the wrapper does not deliver telemetry on its most common paths:

1. **`runDockerodeCommand` never sets `CHALK_DOCKERODE_POST_PUSH_TIMEOUT_MS`**, but `runtime.cjs:181-182` treats it as mandatory and returns `unsupported_no_deadline` when it is absent. As shipped, `chalk __ dockerode_run` is a silent no-op unless the operator happens to export an undocumented variable. This is worth checking against your own validation runs: the pipeline used to exercise this feature set that variable explicitly in its environment, so a green end-to-end result is consistent with the defect rather than evidence against it.
2. **`supportedNode` is an exact-major allowlist** (`22.15+` or `24`), not a minimum. `registerHooks` has been stable since 22.15, but Node 23, 25, 26 and everything later silently skip instrumentation with no diagnostic and no non-zero exit.
3. **The post-push child is spawned with stdout mapped to `ignore`**, discarding both the `chalk-docker-post-push-result/v1` status line and chalk's own report when the configured sink is stdout.

### What the end-to-end evidence does and does not show

I verified the retained artifacts from the CI validation run rather than relying on the summary:

- An **unwrapped** dockerode push produced no chalk report at all; the **wrapped** push produced exactly one push report with `_OP_CHALK_COUNT: 1` and `_OP_UNMARKED_COUNT: 0`, carrying registry, repository, tag, and both digest forms plus CI metadata.
- **Subject-digest agreement across four independent places**: the report's `_REPO_DIGESTS`, the attestation fallback tag name, the in-toto statement subject, and the referring descriptor.
- The attestation was **absent on six probes** (referrers API, bare fallback tag, and legacy `.att`, each immediately and after a settle interval) taken *before* the push, and present afterwards.
- **Both arms pushed byte-identical images**, so post-push collection does not mutate the artifact.
- Exactly **one handled Engine push request per arm**.
- Post-fix, a live dockerode 5.0.1 push resolved in the registry to the same digest the post-hook reported (`sha256:771294d0c401e31f…`, confirmed by querying the registry directly), left **zero** loader directories behind, and preserved a non-zero child exit code.

That is a genuinely good result for the mechanism. The caveat is finding 1: that validation ran with the deadline variable supplied by the harness, so it exercises the instrumented path rather than the default path a user gets.

### Note on verification scope

Standard `cosign verify-attestation` was **not** shown to consume the published attestation; the signature check that passed was a detached DSSE verification with the publisher's key, via a purpose-built chain validator. Treat attestation interoperability as unproven here.

### Confidence

Eight of the ten findings were independently re-verified and are stated as confirmed. Two are **plausible** rather than confirmed, and are flagged as such so they can be triaged accordingly:

- `runtime.cjs:275` — the promise path rejecting a push that actually succeeded. The asymmetry with the callback branch is real and visible in the code; what is not directly demonstrated is a concrete input that makes `finishBuffered` throw in practice.
- `runner.nim:64` — loader-directory removal while descendant processes still reference it. The ordering is real; it requires the wrapped command to leave a Node process running past its own exit, which not every publish flow does.

The other eight should be treated as defects, not as candidates.

src/dockerode_post_push/runtime.cjs
```diff
@@ -0,0 +1,384 @@
...
+    let stderrObserved = false;
+    let killTimer = null;
+    const detached = process.platform !== 'win32';
+    const chalkArgs = process.env.CHALK_DOCKERODE_NO_EXTERNAL_CONFIG === '1'
+      ? ['--no-use-external-config', '__', 'docker_post_push']
+      : ['__', 'docker_post_push'];
+    const child = spawn(chalk, chalkArgs, {
+      detached,
+      env: process.env,
+      stdio: ['pipe', 'ignore', 'pipe'],
```
DavidsonErickson: **Post-push child's stdout is discarded, taking the result contract with it.**

`stdio: ['pipe','ignore','pipe']` maps the child's fd 1 to `/dev/null`. `runCmdDockerPostPush` writes the `chalk-docker-post-push-result/v1` record with `stdout.writeLine`, and `doReporting("report")` emits the push report — on a config whose report sink is stdout, the telemetry is collected and then thrown away.

It also makes failures undiagnosable: the caller sees only `posthook_failed` with a bare exit code, so the documented status values (`digest_mismatch_or_collection_failed`, `unsupported_auth`, `unsupported_schema`) never reach anyone. Capturing stdout and surfacing `status` in the diagnostic would fix both.

src/docker/push.nim
```diff
@@ -19,10 +20,55 @@ import "."/[
...
+
+  forceChalkKeys(["DOCKER_PLATFORM"])
+  let chalk = chalkOpt.get()
+  chalk.withErrorContext():
+    if not chalk.isChalked():
+      warn("docker: " & chalk.name & " is not chalked. reporting will be limited")
+      suspendChalkCollectionFor("attestation")
+      suspendChalkCollectionFor("docker")
+
+    loginToRegistries()
```
DavidsonErickson: **`loginToRegistries()` discards the SDK credentials installed moments earlier.**

`runCmdDockerPostPush` sets `dockerAuth = some({auths:{...}})` from the SDK payload; `dockerPostPush` then calls `loginToRegistries()`, whose success path calls `resetDockerAuthConfig()` (`login.nim:83`), setting it back to `none`.

Every later registry read in `collectRunTimeArtifactInfo` / `completeBuildContextUploads` then falls back to `~/.docker/config.json`, finds no credential for the SDK's registry, and 401s — so the push exits `digest_mismatch_or_collection_failed`. This only bites users who configured `login_method: "get"`, which makes it the harder case to debug.

src/commands/cmd_internal.nim
```diff
@@ -6,18 +6,130 @@
...
+  try:
+    putEnv("DOCKER_HOST", "unix://" & socketPath)
+    delEnv("DOCKER_CONTEXT")
+
+    if hasAuth:
+      var dockerConfig = %*{"auths": {}}
+      dockerConfig["auths"][server] = %*{
+        "auth": encode(username & ":" & password),
+      }
+      setDockerAuthConfig(dockerConfig)
```
DavidsonErickson: **Whole-config replacement evicts credentials for every other registry.**

`getDockerAuthConfig()` (`exe.nim:312`) only reads `~/.docker/config.json` when `dockerAuth.isNone()`. Installing a config containing solely the SDK-supplied registry permanently suppresses that lazy load for the rest of the process, so mirror pushes, base-image manifest fetches and attestation reads against any other registry lose credentials that exist on disk.

Merging the SDK entry into the file-derived config, rather than replacing it, keeps both.

src/commands/cmd_internal.nim
```diff
@@ -6,18 +6,130 @@
...
+    repository = payload{"repository"}.getStr()
+    tag = payload{"tag"}.getStr()
+    digest = payload{"digest"}.getStr()
+    socketPath = payload{"socketPath"}.getStr()
+    if operationId == "" or repository == "" or tag == "" or
+       not digest.startsWith("sha256:") or digest.len != 71 or
+       not digest[7 .. ^1].allCharsInSet({'0' .. '9', 'a' .. 'f'}) or
+       '\n' in repository or '\r' in repository or
+       '\n' in tag or '\r' in tag or
+       socketPath notin ["/var/run/docker.sock", getHomeDir() / ".docker/run/docker.sock"]:
```
DavidsonErickson: **Socket allowlist diverges from the Node side when `HOME` is unset.**

This builds the rootless path from Nim's `getHomeDir()`, while dockerode resolves it via `os.homedir()`. With `HOME` unset — common in CI containers and systemd units — `os.homedir()` falls back to `getpwuid` (`/root/.docker/run/docker.sock`) while `getHomeDir()` returns `"/"`, making the allowlist entry `/.docker/run/docker.sock`.

The `socketPath notin [...]` test then rejects a valid rootless push with exit 2 `unsupported_input`, and the Node side records only `posthook_failed`.

src/dockerode_post_push/runtime.cts
```diff
@@ -0,0 +1,384 @@
...
+  source.on('data', (chunk) => {
+    tracker.write(chunk);
+  });
+  source.once('error', (error) => {
+    sourceFailed = true;
+    ended = true;
+    proxy.destroy(error);
+  });
+  source.pipe(proxy, { end: false });
+  source.once('end', async () => {
```
DavidsonErickson, line 490: **Push completion is withheld from the caller until the post-push child exits.**

The source stream's `'end'` handler awaits `invokePostPush()` before calling `proxy.end()`, so `followProgress(stream, onFinished)` and `await finished(stream)` fire when chalk finishes collecting, not when Docker finishes pushing.

With the deadline ceiling at 5 minutes (`runtime.cjs:16`) and a chalk child blocked on an unreachable registry, a 3-second push presents to the application as a 5-minute hang — long enough to blow a CI step timeout and fail a publish that actually succeeded. Signalling `end` on push completion and letting collection finish out-of-band would decouple them.

src/dockerode_post_push/runtime.cjs
```diff
@@ -0,0 +1,384 @@
...
+  if (!terminal.digest) {
+    diagnostic(terminal.overflow ? 'terminal_frame_too_large' : 'missing_terminal_digest');
+    return value;
+  }
+  await invokePostPush(operation, terminal.digest);
+  return value;
+}
+
+function finishStream(source, operationPromise) {
+  const proxy = new PassThrough();
```
DavidsonErickson: **`destroy()` on the returned stream no longer aborts the push.**

`finishStream` hands back a new `PassThrough` rather than the docker-modem response stream. Unpatched, `stream.destroy()` tears down the underlying request; patched, destroying `proxy` does not propagate to `source`, so the push runs to completion, `source`'s `'end'` handler still fires, and `invokePostPush` still reports a push the application believes it cancelled.

The unconditional `proxy.end()` then also runs against an already-destroyed writable with no `'error'` listener attached.

src/dockerode_post_push/runtime.cjs
```diff
@@ -0,0 +1,384 @@
...
+        diagnostic('posthook_complete', { operationId: payload.operationId, digest });
+      }
+      settle();
+    });
+    child.stdin.on('error', () => {});
+    child.stdin.end(JSON.stringify(payload));
+  });
+}
+
+async function finishBuffered(value, operationPromise) {
```
DavidsonErickson: **Promise path can reject a push that actually succeeded.**

`finishBuffered` awaits `operationPromise` and `invokePostPush` with no `try`/`catch`, and the promise branch returns it directly. A synchronous `spawn()` throw for a bad `CHALK_DOCKERODE_CHALK`, or a rejection from `supportedOperation`, surfaces as a rejected `push()` promise — so `await image.push(...)` throws and the caller treats a completed registry push as failed, aborting or retrying the release.

The callback branch already guards this exact case; the promise branch should match it.

src/dockerode_post_push/runner.nim
```diff
@@ -0,0 +1,66 @@
...
+    if hadChalk:
+      putEnv("CHALK_DOCKERODE_CHALK", priorChalk)
+    else:
+      delEnv("CHALK_DOCKERODE_CHALK")
+    if hadNoExternal:
+      putEnv("CHALK_DOCKERODE_NO_EXTERNAL_CONFIG", priorNoExternal)
+    else:
+      delEnv("CHALK_DOCKERODE_NO_EXTERNAL_CONFIG")
+    try:
+      removeDir(loaderDir)
```
DavidsonErickson, line 52: **Loader directory is removed while surviving child processes still reference it.**

The `finally` deletes the loader dir as soon as the wrapped command returns, but `NODE_OPTIONS=--require=<that path>` has already been inherited by anything the command left running — a watcher, a registry proxy, an npm daemon.

The parent restores its own environment, but the surviving tree still points at the deleted path, so the next Node process those descendants spawn aborts at startup with `Cannot find module` and a non-zero exit. Keying cleanup to the process tree, or using a stable per-run path reaped later, avoids it.

src/dockerode_post_push/runtime.cjs
DavidsonErickson: Fixed in 06c1baf. Destroying the returned PassThrough now destroys the underlying Engine stream, records the operation as cancelled, and suppresses post-push processing. Node coverage verifies source destruction and absence of a post-hook payload.

src/docker/push.nim
DavidsonErickson: Fixed in 06c1baf together with the related auth-replacement finding. Dockerode credentials are now retained as a process-local overlay and reapplied after resetDockerAuthConfig reloads credentials following docker login. Native coverage verifies overlay persistence across a reset.

src/dockerode_post_push/runtime.cjs
DavidsonErickson: Fixed in 06c1baf. The buffered Promise instrumentation path now catches operation and post-hook exceptions, emits a diagnostic, and returns the original Dockerode value. A regression test forces an instrumentation-only rejection and verifies fail-open behavior.

src/commands/cmd_internal.nim
DavidsonErickson: Fixed in 06c1baf together with the credential-reset finding. The SDK registry entry is merged over a deep copy of the ambient Docker auth configuration, preserving other registries and top-level credential settings while letting the explicit SDK entry win on a duplicate key.

src/commands/cmd_internal.nim
DavidsonErickson: Fixed in 06c1baf. Nim socket validation now matches Node homedir resolution on POSIX: HOME is preferred, then getpwuid(getuid()) is used when HOME is absent. Native coverage exercises both paths.

src/dockerode_post_push/runner.nim
DavidsonErickson: Fixed in 06c1baf. The wrapper now supplies the existing five-minute ceiling when the deadline variable is absent, preserves explicit caller overrides, and restores the prior environment after the wrapped command. Native coverage verifies the default, override, and restoration paths.

src/dockerode_post_push/runtime.cjs
DavidsonErickson: Fixed in 06c1baf. The post-hook stdout is now captured with a bounded line buffer: the versioned result record for the matching operation is consumed into diagnostics, while all other stdout is forwarded so configured report sinks remain observable. Node coverage verifies report forwarding and result filtering.

src/dockerode_post_push/register.cjs
DavidsonErickson: Resolved in signed commit b0850e95d88a843f221e4b0e46ad4c6a7d1c9f99. The embedded policy supports Node 22.15+, 24, and 26; keeps Node <=21 and known-EOL 23/25 disabled even with a mocked capability; enables majors >26 only in best-effort mode when module.registerHooks() exists; and fails open when the capability is missing. Hosted matrix jobs for exact 22.15.0, latest 22, latest 24, and latest 26 all passed the complete Dockerode suite: https://github.com/crashappsec/chalk/actions/runs/31435324007. The full Node 26 GitLab-to-Frank acceptance, including matching digest and exact build/push/exec/heartbeat SQL rows, is retained at https://gitlab.com/crashappsec/chalk-dockerode-trial-shim/-/blob/46bd4559a54b107984bc132da2f61502b6bd3643/evidence/post-pr-frank-guppy-validation-20260810.md#node-26-support-follow-up.

src/dockerode_post_push/runner.nim
DavidsonErickson: We are resolving this by making the lifecycle an explicit support boundary rather than extending loader lifetime. `dockerode_run` supports a foreground publisher tree only: every instrumented Node descendant must finish before the wrapped command returns; backgrounded or daemonized descendants are unsupported. This is now stated in the built-in help, implementation contract, changelog, PR description, and public docs in c8d2fb19385398524b5f584af938fd6fc16ac320 and crashappsec/chalkproject.io@924d20e6a3b91a737e37051509b0e22c2e7426ab. It matches the motivating publisher, which awaits each Dockerode `publish()` and its `followProgress` completion, and the foreground GitLab acceptance invocation. Supporting surviving descendants would require a separate stable-loader/reaping design and remains out of scope.

src/docker/repo_digest.nim
```diff
@@ -0,0 +1,23 @@
+##
+## Copyright (c) 2026, Crash Override, Inc.
+##
+## This file is part of Chalk
+## (see https://crashoverride.com/docs/chalk)
+##
+
+import ./ids
+
+proc repoDigestMatches*(image, expectedDigest: string,
```
miki725: chalk has docker repo type as well as image digest struct. for both most utils are in ids.nim

src/docker/repo_digest.nim
```diff
@@ -0,0 +1,23 @@
...
+                        repoDigests: seq[string]): bool =
+  let
+    expected         = parseImage(image)
+    expectedHash     = expectedDigest.extractDockerHash()
+    expectedRegistry = expected.registry
+    expectedName     = expected.name
+  for value in repoDigests:
+    let candidate = parseImage(value)
+    if candidate.registry == expectedRegistry and
```
miki725: the comparison should happen on the docker image struct via the tuple comparison. thats how rest of chalk uses it

src/docker/exe.nim
```diff
@@ -308,7 +308,22 @@ proc getFrontendVersion*(ctx: DockerInvocation): Option[Version] =
...
+
+proc mergeDockerAuthConfig*(base, overlay: JsonNode): JsonNode =
+  ## Merge process-local credentials over the ambient Docker configuration.
+  ## The overlay wins for duplicate registry keys and survives cache resets.
+  result = if base == nil or base.kind != JObject: newJObject() else: base.deepCopy()
+  if overlay == nil or overlay.kind != JObject or
+     overlay{"auths"} == nil or overlay{"auths"}.kind != JObject:
+    return
+  if result{"auths"} == nil or result{"auths"}.kind != JObject:
+    result["auths"] = newJObject()
```
miki725: there are bunch of utils in json.nim for defaulting asserting json structs which chalk uses in most places

```nim
let foo =
  try:
    base.default(newJObject()).assertIs(JObject)
  except:
    newJObject() 
```

src/docker/exe.nim
```diff
@@ -308,7 +308,22 @@ proc getFrontendVersion*(ctx: DockerInvocation): Option[Version] =
...
+  for registry, auth in overlay{"auths"}.pairs():
+    result["auths"][registry] = auth.deepCopy()
```
miki725: there is `.update()`

src/docker/exe.nim
```diff
@@ -321,11 +336,21 @@ proc getDockerAuthConfig*(): JsonNode =
...
     except:
       trace("docker: could not read docker auth config file " & path & " due to: " & getCurrentExceptionMsg())
+    if dockerAuthOverlay.isSome():
+      dockerAuth = some(mergeDockerAuthConfig(dockerAuth.get(), dockerAuthOverlay.get()))
   return dockerAuth.get()
 
 proc resetDockerAuthConfig*() =
   dockerAuth = none(JsonNode)
 
+proc setDockerAuthConfig*(config: JsonNode) =
```
miki725: this doesnt actually set the auth which is used by the docker registry apis. it adds overlay config. function should be renamed

src/commands/cmd_internal.nim
```diff
@@ -6,18 +6,135 @@
...
+    username    = ""
+    password    = ""
+    server      = ""
+    hasAuth     = false
+  try:
+    let payloadInput = readDockerPostPushPayload(stdin)
+    if payloadInput.len > dockerPostPushMaxPayloadBytes:
+      postPushResult("unsupported_input", operationId)
+      quitChalk(2)
+    let payload = parseJson(payloadInput)
```
miki725: can we use chalk json utils for validating json objects

it makes it more chainable functional vs lots if statements

src/docker/push.nim
```diff
@@ -19,10 +20,55 @@ import "."/[
...
+  ## never invokes `docker push`, including configured mirror pushes.
+  let local = inspectImageJson(image)
+  if not repoDigestMatches(image, expectedDigest, local{"RepoDigests"}.getStrElems()):
+    error("docker: post-push digest does not match local RepoDigests for " & image)
+    return false
+
+  let chalkOpt = scanImage(image, fromManifest = false)
+  if chalkOpt.isNone():
+    error("docker: " & image & " is not found; post-push collection skipped")
+    return false
```
miki725, line 78: this will inspect image multiple times

src/docker/push.nim
```diff
@@ -19,10 +20,55 @@ import "."/[
...
+      suspendChalkCollectionFor("attestation")
+      suspendChalkCollectionFor("docker")
+
+    loginToRegistries()
+    initCollection()
+    chalk.addToAllChalks()
+    chalk.collectedData["_OP_ARTIFACT_CONTEXT"] = pack("push")
+    chalk.collectChalkTimeArtifactInfo()
+    chalk.collectAfterSuccessfulPush()
+  return true
```
miki725: this is heavily duplicating the normal docker push. can we abstract so the actual collection, pushing of attestation, context is sharing the same code across 2 use-cases.

otherwise this can easily diverge

src/dockerode_post_push/runner.nim
```diff
@@ -0,0 +1,85 @@
...
+    else:
+      delEnv("CHALK_DOCKERODE_CHALK")
+    if hadNoExternal:
+      putEnv("CHALK_DOCKERODE_NO_EXTERNAL_CONFIG", priorNoExternal)
+    else:
+      delEnv("CHALK_DOCKERODE_NO_EXTERNAL_CONFIG")
+    if hadTimeout:
+      putEnv("CHALK_DOCKERODE_POST_PUSH_TIMEOUT_MS", priorTimeout)
+    else:
+      delEnv("CHALK_DOCKERODE_POST_PUSH_TIMEOUT_MS")
```
miki725: chalk has a util for env vars which cleanup automatically

https://github.com/crashappsec/chalk/blob/0a38a674a8669888f25183a3c81c3e6133a82daf/src/utils/envvars.nim

src/dockerode_post_push/runner.nim
```diff
@@ -0,0 +1,85 @@
...
+    priorChalk         = getEnv("CHALK_DOCKERODE_CHALK")
+    priorNoExternal    = getEnv("CHALK_DOCKERODE_NO_EXTERNAL_CONFIG")
+    priorTimeout       = getEnv("CHALK_DOCKERODE_POST_PUSH_TIMEOUT_MS")
+    hadNodeOptions     = existsEnv("NODE_OPTIONS")
+    hadChalk           = existsEnv("CHALK_DOCKERODE_CHALK")
+    hadNoExternal      = existsEnv("CHALK_DOCKERODE_NO_EXTERNAL_CONFIG")
+    hadTimeout         = existsEnv("CHALK_DOCKERODE_POST_PUSH_TIMEOUT_MS")
+
+  try:
+    discard chmod(cstring(loaderDir), Mode(0o700))
```
miki725: https://github.com/crashappsec/chalk/blob/0a38a674a8669888f25183a3c81c3e6133a82daf/src/utils/files.nim#L265

src/chalk.nim
```diff
@@ -70,6 +70,8 @@ when isMainModule:
   of "docgen":             runChalkDocGen() # in cmd_help
   of "__.onbuild":         runCmdOnBuild() # in cmd_internal
   of "__.prep_postexec":   runCmdPrepPostExec() # in cmd_internal
+  of "__.docker_post_push": runCmdDockerPostPush() # in cmd_internal
+  of "__.dockerode_run":    runCmdDockerodeRun() # in cmd_internal
```
miki725: this doesnt seem like an internal chalk command. the user is meant to call it whereas `__` internal commands are meant for chalk itself to call. for example with `ONBUILD` from dockerfile

src/dockerode_post_push/register_impl.cts
miki725, line 1: none of the js can be type checked over time here

imo all the node bits should be written in typescript which can provide type-safe compatibility with diff node versions as well as dockerode itself via its own types https://www.npmjs.com/package/@types/dockerode

then we can transpile that into whatever js targets we like during the chalk build step

tests/unit/test_dockerode_post_push.nim
```diff
@@ -0,0 +1,146 @@
...
+  putEnv("TMPDIR", tempRoot)
+  putEnv("NODE_OPTIONS", "--trace-warnings")
+  putEnv("CHALK_DOCKERODE_CHALK", "prior-chalk")
+  putEnv("CHALK_DOCKERODE_NO_EXTERNAL_CONFIG", "prior-no-external")
+  delEnv("CHALK_DOCKERODE_POST_PUSH_TIMEOUT_MS")
+  try:
+    doAssert runDockerodeCommand(
+      "/bin/sh",
+      @["-c", "test \"$CHALK_DOCKERODE_POST_PUSH_TIMEOUT_MS\" = 300000"],
+      noExternalConfig = true,
```
miki725: calling shell scripts doesnt seem like a unit test anymore

should prolly test end-to-end via functional tests

miki725 (changes requested):
mostly looked at the nim code

src/dockerode_post_push/register_impl.cts
DavidsonErickson: Thanks — I agree that type checking would be valuable for the main Node implementation. I’d like to defer that conversion from this review pass because Chalk currently has no root JavaScript build dependency; doing it here would add Node, a package manager, TypeScript, and Dockerode types to the main build in addition to the integration refactor requested in the other threads.

There is one intentional exception either way: `register.cjs` needs to remain a tiny handwritten legacy-CJS bootstrap. Its ES5 syntax and total `try { require("./register_impl.cjs") } catch (_) {}` guard are the fail-open mechanism for old/unsupported runtimes. We reproduced the original unguarded bootstrap crashing on real Node 12 before application startup, while the current bootstrap lets that same application run unchanged. A generated preamble or syntax unsupported by the host could recreate that failure before instrumentation can opt out.

My proposed disposition is therefore: keep `register.cjs` handwritten, address the requested Chalk/Nim integration refactors now, and handle a typed main Node implementation as a follow-up with an explicit build-tooling decision. Is TypeScript a recommendation, or is converting the main implementation an approval blocker for this PR?

src/docker/repo_digest.nim
DavidsonErickson: Accepted and implemented at exact signed head `1d4988c`. Pushed-digest matching now lives in `src/docker/ids.nim` and parses both the requested image and local `RepoDigests` with Chalk’s existing `DockerImage`/`DockerImageRepo` types; the standalone digest module is gone.

src/docker/repo_digest.nim
DavidsonErickson: Accepted with one compatibility qualification. The comparison now uses Chalk image tuples after canonicalizing Docker Hub aliases and supplying the source tag plus expected digest, then compares the complete tuple. This retains tested equivalence among `docker.io`, `index.docker.io`, and `registry-1.docker.io`, while custom-port, empty-list, and mismatch rejection cases remain green.

src/docker/exe.nim
DavidsonErickson: Accepted and implemented. Auth overlay construction now uses Chalk JSON `default` and `assertIs` helpers while preserving the existing deep-copy/no-input-mutation contract. The auth regression battery passes on exact head `1d4988c`.

src/docker/exe.nim
DavidsonErickson: Accepted and implemented. The overlay merge now uses Chalk JSON `update`; nested caller JSON is still deep-copied so the supplied auth object is not mutated.

src/docker/exe.nim
DavidsonErickson: Accepted and implemented. The setter is renamed to `setDockerAuthConfigOverlay`, which describes that it overlays Dockerode credentials onto the existing Docker auth configuration.

src/commands/cmd_internal.nim
DavidsonErickson: Accepted and implemented. The stdin payload parser now validates the object and required fields through Chalk’s chainable JSON helpers (`assertIs`, `assertHasKey`, `assertKind`) rather than the prior manual `if` chain. Malformed-input compiled-binary coverage passes.

src/docker/push.nim
DavidsonErickson: Accepted and implemented. Dockerode completion inspects the local image once; digest validation consumes that JSON and the scan/collection overload reuses the same inspected result. The exact-head native and full pytest suites pass.

src/docker/push.nim
DavidsonErickson: Accepted and implemented. `preparePushCollection`/`completePushCollection` now share auth, collection initialization, artifact context, chalk-time/runtime collection, attestations, and flushing. Only normal `dockerPush` invokes Docker and configured mirrors. The standing private regression normalizes Engine traffic and blocks unless the customer program plus collection issue exactly one image-push request; exact-head job 15942965884 passed it, proving collection added zero pushes.

src/dockerode_post_push/runner.nim
DavidsonErickson: Accepted and implemented. Runner environment mutation now uses Chalk’s `withEnvRestore` and `setEnv`, preserving restoration and timeout precedence. The public-wrapper functional tests and compiled-binary smoke pass.

src/dockerode_post_push/runner.nim
DavidsonErickson: Accepted and implemented. Private loader-directory and file modes now use `chmodFilePermissions`, retaining `0700` for the directory and `0600` for loader files. Permission and cleanup checks pass against the compiled binary.

src/chalk.nim
DavidsonErickson: Accepted and implemented. The user-facing wrapper is now `chalk dockerode -- <command> [args...]`; only the child stdin protocol remains under `chalk __ docker_post_push`. Help, changelog, feature issue, private harness, and draft docs PR 62 are aligned. Exact-head forwarding and exit-status tests pass.

src/dockerode_post_push/register_impl.cts
DavidsonErickson: Deferred for this pass under Eric’s explicit scope override while awaiting your blocker/non-blocker answer; this is not being represented as your approval, and I am leaving this thread unresolved. A main TypeScript conversion would add a root Node/package-manager/TypeScript/Dockerode-types build toolchain where Chalk currently has none. Separately, the 11-line `register.cjs` must remain handwritten legacy CJS: that syntax is the fail-open mechanism that lets unsupported old Node parse the preload and continue. Is TypeScript for the main implementation an approval blocker, or a follow-up recommendation? If it is blocking, I will stop for a fresh scope decision.

tests/unit/test_dockerode_post_push.nim
DavidsonErickson: Accepted and implemented. Process execution, child-status preservation, loader cleanup, permissions, fail-open behavior, and public CLI coverage now live in `tests/functional/test_dockerode.py`; pure parsing, JSON, socket, and digest checks remain Nim units. Exact-head hosted pytest passed 279 tests with 26 skipped, and the four-version Node contract matrix also passed.

src/dockerode_post_push/runner.nim
```diff
@@ -0,0 +1,71 @@
...
+      discard chmodFilePermissions(path, "0600")
+    let envVars = @[
+      setEnv("NODE_OPTIONS", (priorNodeOptions & " --require=\"" & register & "\"").strip()),
+      setEnv("CHALK_DOCKERODE_CHALK", getMyAppPath()),
+      setEnv("CHALK_DOCKERODE_POST_PUSH_TIMEOUT_MS", timeout),
+    ]
+    withEnvRestore(envVars):
+      if noExternalConfig:
+        withEnvRestore(@[setEnv("CHALK_DOCKERODE_NO_EXTERNAL_CONFIG", "1")]):
+          result = runCmdNoOutputCapture(command, commandArgs)
```
DavidsonErickson, line 47: **The wrapped command is never resolved against `PATH`, so the documented usage fails.**

`runCmdNoOutputCapture` -> `runCmdGetEverything` -> nimutils `runCommand`, which ultimately calls `execve`/`execv` on `ctx->cmd` verbatim (`nimutils/c/subproc.c`); there is no `execvp` and no path search. `chalk exec` handles this explicitly via `findAllExePaths(cmdName, cmdPath, usePath)` (`src/commands/cmd_exec.nim:448`), and `dockerPush` only ever passes an already-resolved `getDockerExeLocation()`.

Concrete failure: the usage advertised in `getopts.c4m` and the CHANGELOG — `chalk dockerode -- node app.js` (or `-- npm run publish`) — execs `node` with no `/`, `execv` returns `ENOENT`, and the child prints `node: No such file or directory` and exits 1. The wrapped command never runs. The functional test only passes `sys.executable`, an absolute path, so it does not catch this.

Suggest resolving `command` via the same `findAllExePaths` helper (or `findExePath`) before calling `runCmdNoOutputCapture`, and erroring with a clear message when nothing is found.

src/docker/scan.nim
```diff
@@ -68,7 +73,23 @@ proc scanImage*(name:          string | DockerImage,
   var chalk = newChalk(name     = name,
                        codec    = getPluginByName("docker"),
                        platform = platform)
-  return chalk.scanImage(name, image)
+  return chalk.scanImage(name, image, fromManifest = fromManifest)
```
DavidsonErickson: **This one-line "alignment" silently changes existing `docker push` / `docker pull` behavior.**

Before this PR the public `scanImage` dropped its `fromManifest` argument and always called the private overload with the default `fromManifest = true`. The only two callers that pass `fromManifest = false` are `src/docker/pull.nim:24` and `src/docker/push.nim:89`, so both effectively ran with the registry-manifest fallback enabled. Forwarding the parameter now makes `collectImage(..., collectFromManifest = false)` re-`raise` instead of falling back to `collectImageManifest` (`src/docker/collect.nim:441-446`).

Concrete regression: on a `chalk docker push` where `docker inspect <ref>` fails (image only resolvable in the registry, containerd/multi-arch store quirks, inspect of a digest ref), the exception now propagates, `scanImage` returns `none`, and chalk logs `"is not found. pushing without chalk"` and emits no chalk report at all — where previously it recovered via the registry manifest. Same for `chalk docker pull` (`"cannot be collected"`).

If honoring the flag is the intent, it deserves its own change plus a CHANGELOG entry; the new SDK path already gets local-only collection through the dedicated `contents` overload and does not need this.

src/dockerode_post_push/runtime.cjs
```diff
@@ -0,0 +1,455 @@
...
+    if (callbackFn === undefined) {
+      let result;
+      try {
+        result = original.apply(this, rawArgs);
+      } catch (error) {
+        throw error;
+      }
+      return result.then((value) => {
+        if (normalizedOpts.stream === false) return finishBuffered(value, operationPromise);
+        return finishStream(value, operationPromise);
```
DavidsonErickson: **The Engine stream is replaced by a `PassThrough` even when the operation is classified unsupported, so fail-open does not hold for the streaming path.**

`operationPromise` is not consulted before wrapping: both here and at line 436 the value is unconditionally handed to `finishStream`, which returns a fresh `PassThrough`. The `supported: false` classification is only discovered *inside* `finishStream`, after the substitution.

What dockerode returns for a push is the docker-modem response object (`Modem.prototype.buildRequest` calls back with the raw `http.IncomingMessage`), so callers legitimately see `statusCode`, `headers`, `socket`, `req`, etc. After patching, all of those are gone.

Concrete scenario: on macOS (`unsupported_platform`), against a non-default socket / TCP `DOCKER_HOST` (`unsupported_daemon`), or when the preload is present but `CHALK_DOCKERODE_POST_PUSH_TIMEOUT_MS` is unset (`unsupported_no_deadline`) — i.e. exactly the cases the design promises to be inert — an app doing `const res = await image.push({tag}); if (res.statusCode !== 200) ...` now reads `undefined`, and a diagnostic line is written to the app's stderr on every push. `finishBuffered` correctly returns the original value; the stream path should likewise await the classification and return `value` untouched when `!operation.supported`.

src/dockerode_post_push/register_impl.cts
```diff
@@ -0,0 +1,75 @@
...
+      // Most ancestor directories are not package roots.
+    }
+    dir = path.dirname(dir);
+  }
+  return null;
+}
+
+const nodeSupport = classifyNodeSupport(process.versions.node, registerHooks);
+const nodeSupportDiagnostic = diagnosticForNodeSupport(nodeSupport);
+if (nodeSupportDiagnostic) process.stderr.write(`${nodeSupportDiagnostic}\n`);
```
DavidsonErickson, line 88: **This banner is emitted by every Node process in the wrapped tree, not once per publish.**

`NODE_OPTIONS=--require=<loader>` is inherited by the whole command tree, so the preload runs in every `node`, `npm`, `npx`, and forked child. On any runtime outside the supported set the classification is the same for all of them and each writes a line to the app's stderr.

Concrete scenario: `chalk dockerode -- npm run publish` on Node 20 — npm itself, each lifecycle script, and every tool it spawns prints `[chalk-dockerode] Node 20.x.y is excluded as end-of-life; instrumentation skipped (policy last reviewed ...)`. That is dozens of lines of stderr injected into a command chalk promises not to perturb, and it breaks callers that assert on clean stderr. Consider gating on a debug env var, deduplicating via a `Symbol.for` marker on `globalThis` so only the first process in the tree reports, or routing it to `CHALK_DOCKERODE_LOG` like the other diagnostics.

src/dockerode_post_push/register_impl.cts
miki725: it could easily be orchestrated in a makefile. for linux the typescript env can be all managed in docker. if dont think it makes sense to ship a new feature without using typescript which improves readability and safety a lot

miki725:
how is this meant to be used by chalk users? dockerode users might be running node process directly by jenkins for example. in which case what will be executing the `chalk dockerode -- ...`?

in addition `--require` has compatibility issues with node ESM modules which wont allow the common js module to monkey patch modules.

we already publish a dust npm package as a serverless plugin.

why not publish a dockerode plugin which the customers can manage in their normal deps same as they manage dockerode dep

for the long term I dont think this is correct architecture approach for wrapping node ecosystem as its packaging landscape is so vast, not to even mention things like https://deno.com/ which completely changes the runtime

src/dockerode_post_push/runner.nim
DavidsonErickson: **Withdrawn — this finding is incorrect.** I disproved it by running the built binary rather than reading the call path.

```
$ ./chalk --no-use-external-config dockerode -- node --version
v24.9.0        # exit 0

$ ./chalk --no-use-external-config dockerode -- "$(command -v node)" --version
v24.9.0
```

A bare command name resolves correctly, so the documented `chalk dockerode -- node app.js` / `-- npm run publish` usage works as written. There is no defect here and no change is needed.

The two supporting observations were accurate but did not support the conclusion I drew from them: `cmd_exec.nim` does use `findAllExePaths`, and the functional test does pass an absolute interpreter path rather than a bare name. Neither implies the wrapped command goes unresolved — resolution happens further down than my static read reached.

Apologies for the noise. The other findings on this branch stand as written.

src/dockerode_post_push/register_impl.cts
DavidsonErickson: Implemented at signed exact head `7eb976bc12bad15f222106031036c17cfe83f73f`.

- `runtime`, `register_impl`, `node_support`, the preload bootstrap, and all Node contracts are now strict TypeScript. The main artifacts emit CommonJS for Node 22.15+.
- The typed preload is compiled separately to ES5. Its emitted 428-byte artifact has one guarded `require("./register_impl.cjs")`, no helper/import preamble, and passed missing/corrupt implementation cases on real Node 12.22.12 while preserving application output and exit status.
- Root dependencies are pinned in the lockfile and installed with `npm ci --ignore-scripts`. Chalk release/unit/functional Make entry points transpile before Nim embeds the generated files; generated JavaScript remains ignored and TypeScript/npm are build-time only.
- Strict no-emit checking and two-clean-build determinism passed. All 30 compiled contracts passed on exact Node 22.15.0, latest 22, 24, and 26. Hosted native Linux, full pytest, TypeScript/Node 12, and pre-commit checks are green: [Node/native/TypeScript](https://github.com/crashappsec/chalk/actions/runs/32167541082), [pytest](https://github.com/crashappsec/chalk/actions/runs/32167541081).
- Both exact-head private consumers passed. Canonical `chalk docker build`/`chalk docker push` and Node 26/Dockerode 5.0.1 through public `chalk dockerode` produced matching terminal/local/report digests, runtime exec and heartbeat rows; the Dockerode Engine trace retained the blocking exactly-one-push assertion.

Claude concurred on this exact head after independently reproducing the bootstrap hash and Node policy. Fresh Copilot review [4964798898](https://github.com/crashappsec/chalk/pull/762#pullrequestreview-4964798898) generated no new comments.

This addresses the TypeScript approval blocker, so I am resolving this thread and requesting your rereview.

src/dockerode_post_push/runtime.cts
DavidsonErickson: The approved contract intentionally withholds stream completion only for bounded post-push work. Five minutes remains a proof-only ceiling, not a production timeout. The one-millisecond central-policy run proves that timeout preserves caller success and does not replay the Engine push. Asynchronous post-processing and production timeout selection remain separate work.

src/dockerode_post_push/runner.nim
DavidsonErickson: The foreground process-tree boundary remains explicit. Manual activation removes its temporary loader only after the foreground command returns, so detached descendants are unsupported. Central activation does not use that temporary lifecycle. It pins an immutable release that survives agent rollback.

src/dockerode_post_push/runner.nim
DavidsonErickson: This finding was withdrawn after the built CLI proved that bare node and npm commands resolve through PATH. The current functional suite also exercises npm run publish.

src/docker/scan.nim
DavidsonErickson: The current head preserves the historical public scanImage behavior. The public overload still retains registry-manifest fallback, while the new contents overload isolates the Dockerode handoff. The manifest-fallback guard and focused functional suite pass.

src/dockerode_post_push/register_impl.cts
DavidsonErickson: The concrete Node 20 case no longer applies. Node 20 now uses the supported bounded CommonJS loader and emits no unsupported-version banner. Central activation routes runtime diagnostics to the protected agent-owned file. The unrelated-command regression also compares stdout, stderr, exit status, and filesystem output under the preload.

PR #766, opened 2026-09-07
fix: segfault in certs codec aborting chalked docker builds

maximilianhurl:
## Problem

Chalked Docker builds have been dying at `prep_postexec` and silently falling back to unchalked:

```
#29 10.21 error: pid: 1 - Aborting due to signal: SIGSEGV(11)
#29 ERROR: process "/chalk ... __ prep_postexec" did not complete successfully: exit code: 139
error: docker: retrying without chalk due to: wrapped docker build exited with 1
```

The retry succeeds and pushes, so **CI stays green** while the image carries no chalk mark. Downstream that looks like missing build/push records and no deployment correlation — the symptom that led here.

This is not new and not tied to a release: it reproduces on `1.1.4-dev`, `1.2.0-dev` and `1.2.0`. The same repo on the same chalk version produces both outcomes (47 chalked / 69 not), which is what pointed at input-dependent memory corruption rather than a regression.

## Cause

`prep_postexec` subscans `/` with `scan_codecs = ["certs"]`. With `certs.scan_no_extension` defaulting to true, every extensionless file — including a statically linked Go binary — is fed to `extract_cert_data()`, which falls back to `d2i_X509_bio()` when the PEM read fails. On arbitrary binary content that occasionally yields a structurally valid but malformed certificate.

`src/utils/certs.c:172-174`:

```c
EVP_PKEY *pub    = X509_get_pubkey(cert);
int       keynid = EVP_PKEY_base_id(pub);   // pub is NULL -> SEGV
```

`X509_get_pubkey()` returns NULL when the public key will not parse. Unchecked.

## How it was found

DER-encode a real certificate, flip 1–16 random bytes, drive it through the actual parse loop under ASan+UBSan:

```
==7==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000
    #0 EVP_PKEY_get_base_id (libcrypto.so.3)
    #1 extract_cert_data certs.c:174
```

Crashes within 20,000 mutations on the first seed. The same run flagged a second defect at `certs.c:126`.

## Changes

- **`X509_get_pubkey()` NULL check** — the crash. The record is treated as "not a certificate" instead of dereferencing.
- **`BIO_all()` made binary-safe** — it built its buffer with `strndup()`, which stops at the first NUL, while `total` counted every byte read; later indices then ran past the allocation. Now `calloc` + `memcpy` with bounded trailing writes.
- **`SubjectPublicKeyInfo` instead of `PKCS1`** — PKCS1 is RSA-only, so for EC keys the encoder emitted nothing and a NULL landed mid-`key_value`, truncating both `cleanup_key_value()` and `cstringArrayToSeq()` on the Nim side. Every field after `Serial` was silently lost for EC certificates (39 of 119 in a stock CA bundle).
- **Leaks** — the `X509`, `EVP_PKEY`, `BIGNUM` and encoder context were never released; `cleanup_key_value()` freed the strings but not the arrays; `subject_short`/`issuer_short` were never freed at all. Also an `OPENSSL_free`/`free` mismatch on the serial.
- **Unchecked returns** — `OBJ_nid2ln`/`OBJ_nid2sn`, `ASN1_STRING_length` of 0 (read `data[-1]`, wrote into a zero-byte allocation), and `convert_ASN1TIME`, which returned `-1` even on success and left its buffer uninitialised on early failure.

## Verification

| | before | after |
|---|---|---|
| Fuzz, 160k mutations × 8 seeds | SEGV on seed 1 | clean |
| LeakSanitizer, 119-cert CA bundle | 771,327 bytes / 14,153 allocs | 0 |
| Sweep: full alpine prod rootfs + released chalk 1.2.0 binary + 24MB Go binary | — | clean |

Built with `-fsanitize=address,undefined` against system OpenSSL 3 (one local substitution for the `crypto/x509.h` internal include).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,41 @@
 # Chalk Release Notes
 
+## Unreleased
+
+### Bug Fixes
+
+- Fixed a segfault in the certs codec that aborted chalked Docker builds. The
+  `prep_postexec` step subscans `/` with the certs codec, which feeds every
+  candidate file to `d2i_X509_bio` after the PEM read fails. On arbitrary binary
+  content (a statically linked Go binary, for instance) that occasionally yields
+  a structurally valid but malformed certificate, for which
+  `X509_get_pubkey()` returns `NULL`; the result went straight into
+  `EVP_PKEY_base_id()` and crashed the build with SIGSEGV. Chalk then logged
+  `retrying without chalk` and rebuilt unwrapped, so CI stayed green while the
+  pushed image carried no chalk mark -- no build/push records and no deployment
+  correlation for the affected images.
+
+- Fixed a heap buffer overflow in the certs codec's `BIO_all()`. It built its
+  result with `strndup()`, which stops at the first NUL, while the running
+  `total` counted every byte read. On binary payloads the allocation was
+  therefore smaller than `total` claimed, so the copy loop over-read and the
+  trailing NUL-termination writes ran past the end of the buffer.
+
+- Fixed X.509 metadata being dropped for EC certificates. The public key was
+  encoded with the `PKCS1` structure, which is RSA-only, so for EC keys the
+  encoder produced nothing and a `NULL` landed mid-`key_value`. That truncated
+  both `cleanup_key_value()` and the `cstringArrayToSeq()` on the Nim side, so
+  every field after `Serial` -- key type, signature type, validity, signature
+  and all extensions -- was silently lost. The encoder now uses
+  `SubjectPublicKeyInfo`, which covers every key type.
+
+- Fixed memory leaks in the certs codec. `extract_cert_data()` never released
+  the `X509`, `EVP_PKEY`, `BIGNUM` or encoder context; `cleanup_key_value()`
+  freed the strings but never the arrays holding them; and `subject_short` and
+  `issuer_short` were never freed at all. Scanning a 119-certificate CA bundle
+  leaked 771KB. The serial was also allocated by OpenSSL and released with
+  `free()` rather than `OPENSSL_free()`.
+
 ## 1.2.0
 
 **August 11, 2026**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,12 +14,14 @@
   `retrying without chalk` and rebuilt unwrapped, so CI stayed green while the
   pushed image carried no chalk mark -- no build/push records and no deployment
   correlation for the affected images.
+  ([#766](https://github.com/crashappsec/chalk/pull/766))
 
 - Fixed a heap buffer overflow in the certs codec's `BIO_all()`. It built its
   result with `strndup()`, which stops at the first NUL, while the running
   `total` counted every byte read. On binary payloads the allocation was
   therefore smaller than `total` claimed, so the copy loop over-read and the
   trailing NUL-termination writes ran past the end of the buffer.
+  ([#766](https://github.com/crashappsec/chalk/pull/766))
 
 - Fixed X.509 metadata being dropped for EC certificates. The public key was
   encoded with the `PKCS1` structure, which is RSA-only, so for EC keys the
@@ -28,6 +30,7 @@
   every field after `Serial` -- key type, signature type, validity, signature
   and all extensions -- was silently lost. The encoder now uses
   `SubjectPublicKeyInfo`, which covers every key type.
+  ([#766](https://github.com/crashappsec/chalk/pull/766))
 
 - Fixed memory leaks in the certs codec. `extract_cert_data()` never released
   the `X509`, `EVP_PKEY`, `BIGNUM` or encoder context; `cleanup_key_value()`
@@ -35,6 +38,7 @@
   `issuer_short` were never freed at all. Scanning a 119-certificate CA bundle
   leaked 771KB. The serial was also allocated by OpenSSL and released with
   `free()` rather than `OPENSSL_free()`.
+  ([#766](https://github.com/crashappsec/chalk/pull/766))
 
 ## 1.2.0
 
--- a/src/utils/certs.c
+++ b/src/utils/certs.c
@@ -20,8 +20,20 @@ int
 convert_ASN1TIME(ASN1_TIME *t, char *buf, size_t len)
 {
     int  rc;
-    BIO *b = BIO_new(BIO_s_mem());
-    rc     = ASN1_TIME_print(b, t);
+    BIO *b;
+
+    if (buf == NULL || len == 0) {
+        return -1;
+    }
+    buf[0] = 0;
+    if (t == NULL) {
+        return -1;
+    }
+    b = BIO_new(BIO_s_mem());
+    if (b == NULL) {
+        return -1;
+    }
+    rc = ASN1_TIME_print(b, t);
     if (rc <= 0) {
         BIO_free(b);
         return -1;
@@ -32,17 +44,27 @@ convert_ASN1TIME(ASN1_TIME *t, char *buf, size_t len)
         return -1;
     }
     BIO_free(b);
-    return -1;
+    return 0;
 }
 
 char *
 convert_ASN1STRING(ASN1_BIT_STRING *s)
 {
-    int            l      = ASN1_STRING_length(s);
+    int            l      = (s == NULL) ? 0 : ASN1_STRING_length(s);
+    unsigned char *data;
+    char          *result;
+    char          *cur;
+
+    if (l <= 0) {
+        return calloc(1, 1);
+    }
     // each byte is 2 hex plus colon or last char NULL
-    char          *result = calloc(l * 3, 1);
-    unsigned char *data   = ASN1_STRING_get0_data(s);
-    char          *cur    = result;
+    result = calloc(l * 3, 1);
+    if (result == NULL) {
+        return NULL;
+    }
+    data   = ASN1_STRING_get0_data(s);
+    cur    = result;
 
     for (int i = 0; i < l - 1; i++) {
         sprintf(cur, "%02x:", data[i]);
@@ -72,11 +94,9 @@ convert_NAME(X509_NAME *n, int short_name)
             OBJ_obj2txt(scratch, 200, key, 1);
             key_value[ix++] = strdup(scratch);
         } else {
-            if (short_name) {
-                key_value[ix++] = strdup(OBJ_nid2sn(nid));
-            } else {
-                key_value[ix++] = strdup(OBJ_nid2ln(nid));
-            }
+            const char *nid_name = short_name ? OBJ_nid2sn(nid)
+                                             : OBJ_nid2ln(nid);
+            key_value[ix++] = strdup(nid_name == NULL ? "" : nid_name);
         }
 
         int len = ASN1_STRING_to_UTF8(&utf8, value);
@@ -99,17 +119,31 @@ BIO_all(BIO *bio)
     char    *cur;
     char     scratch[PIPE_BUF];
     int      total;
+    int      lastchar;
 
-    BIO_get_mem_ptr(bio, &bptr);
-    int n = BIO_read(bio, scratch, PIPE_BUF);
-    if (!n) {
+    if (BIO_get_mem_ptr(bio, &bptr) <= 0 || bptr == NULL) {
         return NULL;
     }
-    result = strndup(scratch, n);
+    int n = BIO_read(bio, scratch, PIPE_BUF);
+    if (n <= 0) {
+        return NULL;
+    }
+    // memcpy, not strndup: the payload is binary and may contain NULs, in
+    // which case strndup would allocate only up to the first one while `total`
+    // kept counting every byte read -- so every later index derived from
+    // `total` or bptr->length ran off the end of the allocation.
+    result = calloc((size_t)n + 1, 1);
+    if (result == NULL) {
+        return NULL;
+    }
+    memcpy(result, scratch, n);
     total  = n;
 
     while ((n = BIO_read(bio, scratch, PIPE_BUF)) > 0) {
-        cur = calloc(total + n + 1, 1);
+        cur = calloc((size_t)total + n + 1, 1);
+        if (cur == NULL) {
+            return result;
+        }
         memcpy(cur, result, total);
         memcpy(cur + total, scratch, n);
         free(result);
@@ -117,12 +151,12 @@ BIO_all(BIO *bio)
         total  = total + n;
     }
 
-    int lastchar = bptr->length;
+    lastchar = (int)bptr->length;
 
     // BIO_read sometimes reads more bytes,
     // possibly for not-NULL terminated objects
-    if (bptr->length < total) {
-        result[bptr->length] = 0;
+    if (lastchar >= 0 && lastchar < total) {
+        result[lastchar] = 0;
 
         // remove newlines
         if (lastchar > 1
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,45 @@
...
+  ([#766](https://github.com/crashappsec/chalk/pull/766))
+
+- Fixed X.509 metadata being dropped for EC certificates. The public key was
+  encoded with the `PKCS1` structure, which is RSA-only, so for EC keys the
+  encoder produced nothing and a `NULL` landed mid-`key_value`. That truncated
+  both `cleanup_key_value()` and the `cstringArrayToSeq()` on the Nim side, so
+  every field after `Serial` -- key type, signature type, validity, signature
+  and all extensions -- was silently lost. The encoder now uses
+  `SubjectPublicKeyInfo`, which covers every key type.
+  ([#766](https://github.com/crashappsec/chalk/pull/766))
```
nettrino: Two problems with this entry.

**1. The stated mechanism does not happen.** I traced the pre-fix path: for an EC key the PKCS1 encoder produces an empty mem BIO, and `BIO_read()` on an empty mem BIO returns **-1**, not 0 — so the old `if (!n) return NULL;` guard never fired. `strndup(scratch, (size_t)-1)` then returns a non-NULL string, and `bptr->length (0) < (size_t)total (-1)` evaluates TRUE, so `result[0] = 0`. The result is an **empty string, not `NULL`**. Nothing lands mid-`key_value` and nothing is truncated.

The repo proves it independently: `test_cert` **on `main`** already asserts `_X509_SIGNATURE` for the COMODO **ECC** certificate, and `Signature` sits *after* `Key` in `key_value`. That test passes today, so "every field after `Serial` was silently lost" cannot be right.

The actual symptom is narrower: `_X509_KEY` was empty for EC certificates, and `setIfNeeded` drops empty strings, so the key was simply absent from the report. Everything else was intact.

The real pre-fix defect on this path is arguably worse than what's written and worth stating instead: `strndup(scratch, (size_t)-1)` runs `strnlen` over a 4096-byte **uninitialized stack buffer** with no bound — an out-of-bounds read that happens to be survivable.

**2. A breaking data change is undocumented.** Switching PKCS1 → `SubjectPublicKeyInfo` also changes `_X509_KEY` for **RSA** certificates, from `-----BEGIN RSA PUBLIC KEY-----` to `-----BEGIN PUBLIC KEY-----`. Verified against OpenSSL 3.6.3. Anything downstream that parses that field breaks. This belongs under a Breaking Changes / Changed heading, not folded into an EC bug fix.

src/utils/certs.c
```diff
@@ -56,36 +85,48 @@ convert_ASN1STRING(ASN1_BIT_STRING *s)
...
-                key_value[ix++] = strdup(OBJ_nid2ln(nid));
-            }
+            const char *nid_name = short_name ? OBJ_nid2sn(nid)
+                                             : OBJ_nid2ln(nid);
+            key_value[ix] = strdup(nid_name == NULL ? "" : nid_name);
+        }
+        if (key_value[ix] == NULL) {
+            cleanup_key_value(key_value);
+            return NULL;
         }
```
nettrino, line 127: This introduces a new `NULL` return from `convert_NAME()`, and the Nim side cannot take it. `extract_cert_data()` assigns the result straight into `Cert.subject` / `subject_short` / `issuer` / `issuer_short`, and `src/plugins/codecCerts.nim:51` calls `cstringArrayToSeq(t)` on each, which dereferences `a[0]` without a NULL check — so a `NULL` here segfaults in exactly the place this PR is trying to stop segfaulting.

Allocation-failure-only, so low likelihood, but the fix is one line: return an empty terminated array rather than `NULL`.

```c
cleanup_key_value(key_value);
return calloc(1, sizeof(*key_value));
```

(Same at the second guard below.)

src/utils/certs.c
```diff
@@ -170,18 +197,35 @@ extract_cert_data(BIO *fdb)
     char              scratch[2000];
     int               version    = ((int)X509_get_version(cert)) + 1;
     EVP_PKEY         *pub        = X509_get_pubkey(cert);
+    if (pub == NULL) {
+        // Unparseable/unsupported public key. Treat the whole record as not a
+        // certificate rather than dereferencing NULL below.
+        X509_free(cert);
+        return NULL;
+    }
```
nettrino, line 217: Correct fix for the crash — confirmed this is reachable with `cert != NULL` and `pub == NULL`.

One consequence worth recording: `findCerts` in `src/plugins/codecCerts.nim:66-67` `break`s on a `NULL` return, so a single record with an unloadable public key now silently drops **every remaining certificate in the same file**. Irrelevant for the reported case (a Go binary is one record), but for a PEM bundle containing one cert whose algorithm the active providers can't load, everything after it disappears with no diagnostic. The iterator can't distinguish "end of input" from "bad record" as written.

Not a blocker — it segfaulted before — but a `trace()` on this path, or a sentinel that lets the loop continue, would make the drop visible.

src/utils/certs.c
```diff
@@ -208,13 +255,25 @@ extract_cert_data(BIO *fdb)
...
 
     int ix              = 0;
     key_value[ix++]     = strdup("Serial");
     key_value[ix++]     = serial;
     key_value[ix++]     = strdup("Key");
-    key_value[ix++]     = key_contents;
+    key_value[ix++]     = (key_contents == NULL) ? strdup("") : key_contents;
```
nettrino, line 288: Good catch guarding this one, but it's the only guarded insert. Every other `key_value[ix++] = strdup(...)` in this function is unchecked, as is `convert_ASN1STRING(sig)` — which now has two `NULL` return paths of its own (the `SIZE_MAX / 3` check and the `calloc` failure).

A `NULL` in any of those slots reproduces precisely the mid-array truncation this PR sets out to eliminate: `cleanup_key_value()` stops early and leaks the tail, and `cstringArrayToSeq()` silently drops every field after it. Allocation-failure-only, so not urgent, but if it's worth guarding here it's worth guarding uniformly.

src/utils/certs.c
```diff
@@ -96,46 +137,32 @@ BIO_all(BIO *bio)
...
+        if (n <= 0) {
+            free(result);
+            return NULL;
         }
+        total += (size_t)n;
     }
+    result[total] = 0;
 
     return result;
 }
```
nettrino, line 180: This rewrite is a clear improvement and I verified it is behavior-preserving.

Compiled the old and new implementations side by side against a real mem BIO (PEM public key, plus several `X509V3_EXT_print` outputs) and the results are byte-identical, trailing newline included. The old newline-stripping branch was unreachable: `BIO_get_mem_ptr()` returns `bbm->buf`, while `mem_read()` decrements `bbm->readp`, so `bptr->length` never falls below `total` and the branch never executed. Dropping it loses nothing.

Sizing off `bptr->length` up front is also correct here — `mem_buf_sync()` runs inside the `BIO_get_mem_ptr()` ctrl, so `length` is the full unread remainder at call time, and nothing reads from these BIOs beforehand.

tests/functional/test_cert.py
```diff
@@ -58,3 +79,18 @@ def test_cert(
...
+    tmp_data_dir: Path,
+):
+    malformed_cert = tmp_data_dir / "malformed.der"
+    malformed_cert.write_bytes(malformed_public_key_der(server_cert.read_bytes()))
+    insert = chalk.extract(
+        config=CONFIGS / "certs.c4m",
+        artifact=malformed_cert,
+        expecting_chalkmarks=False,
+    )
+    assert "_CHALKS" not in insert.report
```
nettrino, line 96: Verified this test is real rather than vacuous. My first concern was that corrupting the key would make `d2i_X509_bio()` fail outright, in which case `extract_cert_data()` returns early and the test would pass on unpatched code too.

Built a probe against OpenSSL 3.6.3 using this exact mutation:

```
d2i_X509_bio    -> OK (cert parsed)
X509_get_pubkey -> NULL  <-- crash path reached
```

The `X509_PUBKEY` decode failure is soft, so the certificate parses and the `NULL` surfaces at `X509_get_pubkey()` — the guard added in this PR. Good regression test.

One gap: it only asserts the process survives. It doesn't cover the bundle-truncation behavior noted on `src/utils/certs.c:199` — a PEM file with a good cert *after* the malformed one would show that the remaining records are dropped.

nettrino (commented):
Reviewed the C changes and independently re-derived the claims against OpenSSL 3.6.3. The core fix is correct and the regression test is genuine — I verified that the test's mutation (flipping the public key's `SEQUENCE` tag `0x30`→`0x31`) makes `d2i_X509_bio()` succeed while `X509_get_pubkey()` returns `NULL`, so it really does exercise the crash path rather than passing trivially on a `d2i` failure.

Also confirmed the `BIO_all()` rewrite is output-identical to the old one: the old trailing-newline-stripping branch was dead, because `BIO_get_mem_ptr()` hands back `bbm->buf` while `mem_read()` decrements `bbm->readp`, so `bptr->length < total` never held. Ran both versions side by side over a PEM public key and several extension values — byte-identical. No silent format regression there.

Two things in the writeup are factually wrong and one behavior change is undocumented; details inline. Neither blocks the fix, but the changelog text ships to users.

### Not inline-able (outside the diff hunks)

**`src/utils/certs.c:307-308`** — the extension loop still does `strdup(OBJ_nid2ln(nid))` with no NULL guard, while `convert_NAME()` got exactly that guard in this PR. Same for `unsigned nid` on line 300 vs. the `int nid` you corrected on line 100. Safe in practice (a nid from `OBJ_obj2nid()` always resolves), but inconsistent with the hardening standard the rest of the PR sets.

**`src/utils/certs.c:296-298`** — with `BIO_all()` no longer touching the caller's `bptr`, the `BUF_MEM *bptr = NULL; BIO_get_mem_ptr(ext_bio, &bptr);` pair in the extension loop is now fully dead. Worth deleting while you're here.

### Follow-up, not this PR

The root cause named in the description is still live: `certs.scan_no_extension` defaults to `true` (`src/configs/chalk.c42spec:3152`), so `prep_postexec` feeds every extensionless file on `/` — including multi-MB Go binaries — to `d2i_X509_bio()`. Hardening the parser stops the crash; it does not stop chalk from attempting to cert-parse the whole rootfs on every build.

CHANGELOG.md
maximilianhurl: Both correct, and both fixed in 37ed84b.

I compiled the pre-fix `BIO_all()` verbatim against OpenSSL 3.6.3 to check this rather than reason about it:

| case | encoder output | old `BIO_all()` returns |
|---|---|---|
| EC P-256 + `PKCS1` | 0 bytes (`to_bio` -> 0) | `""` — non-NULL, empty |
| EC P-256 + `SubjectPublicKeyInfo` | 178 B | `-----BEGIN PUBLIC KEY-----` |
| RSA 2048 + `PKCS1` | 426 B | `-----BEGIN RSA PUBLIC KEY-----` |
| RSA 2048 + `SubjectPublicKeyInfo` | 451 B | `-----BEGIN PUBLIC KEY-----` |

Confirms your trace exactly: `BIO_read` returns -1, the `if (!n)` guard never fires, and `bptr->length (0) < (size_t)total (-1)` is TRUE via the signed->unsigned conversion, so `result[0] = 0`. Empty string, never NULL. The entry now describes the real symptom — `_X509_KEY` absent for EC certs because `setIfNeeded` -> `setIfNotEmpty` drops empty values — and says explicitly that no other field was affected.

The RSA header change is now its own **Breaking Changes** entry at the top of Unreleased.

I also tightened the `BIO_all()` entry while in there. Its "the trailing NUL-termination writes ran past the end of the buffer" half was not demonstrated either — as you note, that branch is dead for mem BIOs. The reachable defect is the `memcpy(cur, result, total)` over-read after `strndup` stops at a NUL, so that is what it says now, plus the unbounded `strndup(scratch, (size_t)-1)` over the uninitialised stack buffer you identified.

src/utils/certs.c
maximilianhurl: Fixed in 37ed84b — this was the strongest finding in the review. Agreed the PR was introducing a segfault into the function that exists to remove one.

Took your fix, with two additions:

- Routed it through a small `empty_key_value()` helper so all three exits are consistent, including the initial `calloc` failure which had the same problem.
- Added a `count < 0` guard. `X509_NAME_entry_count()` returns 0 for a NULL name in practice, but `(size_t)count * 2 + 1` on a negative count would be catastrophic, and it is free to rule out.

One caveat worth stating: `empty_key_value()` is itself a `calloc` and can return NULL under genuine memory pressure. The window is much narrower — 8 bytes rather than an arbitrary-sized array — but it is not zero, so I closed it properly on the Nim side too: `toTable()` now returns early on a nil `cstringArray`. Note the obvious-looking alternative does not work: a `static char *const empty[] = {NULL}` would be passed to `cleanup_key_value()`, which ends in `free(kv)`.

src/utils/certs.c
maximilianhurl: Thanks for checking this one independently — matches what I had. Your point about the old newline-stripping branch being dead is the more useful half, and it made me re-read the changelog entry for `BIO_all()`, which was overclaiming as a result: it asserted the trailing NUL-termination writes ran past the end of the buffer, which is not reachable if that branch never executes. Corrected in 37ed84b to the defect that is actually reachable (the `memcpy` over-read), plus the unbounded `strndup` you flagged on the other thread.

src/utils/certs.c
maximilianhurl: Deliberate rather than an oversight, though the PR should have said so.

The asymmetry tracks reachability, not tidiness. `key_contents` is NULL on a **non-OOM** path: `OSSL_ENCODER_CTX_new_for_pkey()` can return NULL for a key type with no encoder, and `BIO_all()` returns NULL on an empty BIO — which is exactly what the EC/PKCS1 case did before this branch. That one is a real input-dependent path, so it is guarded.

Every other slot you list is NULL only if `strdup`/`calloc` fails. `convert_ASN1STRING` included: its `SIZE_MAX / 3` branch is unreachable on 64-bit, since `l` comes from `ASN1_STRING_length()` and is bounded by `INT_MAX`, so `calloc` failure is the only way it returns NULL.

I would rather not guard all fourteen inserts uniformly, because it implies those paths are reachable and adds a fair amount of noise to make the OOM case marginally better-behaved when the process is already failing. Happy to be talked out of it if you would rather have the consistency.

src/utils/certs.c
maximilianhurl: Confirmed, and worse than theoretical — I built a fixture and measured it.

Same mutation the test uses, as a two-record PEM:

| input | certs parsed |
|---|---|
| good cert alone | 1 |
| malformed alone | 0, no crash |
| malformed **then** good | **0** |

So a good certificate following a bad record is silently dropped, exactly as you describe.

Not fixing it here, for the reason you identified: the iterator genuinely cannot distinguish "end of input" from "bad record", and flipping `break` to `continue` in `findCerts` would spin forever at EOF. It needs a real sentinel through the C boundary, which is more than this PR should carry — agreed it is follow-up work.

tests/functional/test_cert.py
maximilianhurl: Agreed on the gap, and I verified the drop rather than leaving it hypothetical — details and numbers on the `certs.c:199` thread.

Deliberately not adding that test here. A test today could only assert the truncation, which locks in behaviour the follow-up fix would invert. Better to add it alongside that fix, where it asserts the bundle survives.

maximilianhurl:
Thanks for the review — the two changelog corrections in particular were worth catching, since that text ships.

Pushed as 37ed84b.

## Fixed

- **Changelog, EC entry.** The stated mechanism does not occur; corrected to the real symptom. Verified your trace against OpenSSL 3.6.3 by compiling the pre-fix `BIO_all()` verbatim — table on the changelog thread.
- **Changelog, breaking change.** PKCS1 -> `SubjectPublicKeyInfo` also moves RSA from `BEGIN RSA PUBLIC KEY` to `BEGIN PUBLIC KEY`. Now its own Breaking Changes entry.
- **Changelog, `BIO_all()` entry.** Your "that branch is dead" point applies to this entry too — it claimed trailing writes ran past the end of the buffer, which is unreachable if the branch never executes. Narrowed to the reachable `memcpy` over-read, plus the unbounded `strndup` over the uninitialised stack buffer.
- **`convert_NAME()` NULL return.** The strongest finding — the PR was introducing a segfault into the function meant to remove one. All three exits now return an empty terminated array, plus a `count < 0` guard.
- **Your three non-inline-able items.** Dead `BUF_MEM` pair deleted, `OBJ_nid2ln()` guarded, `unsigned nid` -> `int`.

While fixing the third of those I introduced a new problem and caught it on a second pass: falling back to `""` for a NULL `OBJ_nid2ln()` puts an empty key into `key_value`, and `certsCallback` indexes `k[0]` unconditionally (`src/plugins/codecCerts.nim:246`) — `IndexDefect` with bound checks on. It now falls back to the raw OID text, as the `NID_undef` branch does, and `extname` is initialised before `OBJ_obj2txt()`. The `k[0]` index is length-guarded as well.

`toTable()` also now guards a nil `cstringArray`, which closes the residual case where the empty-array helper is itself defeated by allocation failure.

## Responding, not fixing

- **Uniform guarding of the other inserts.** Deliberate, and the PR should have said so: `key_contents` is NULL on a reachable non-OOM path, everything else only under allocation failure. Reasoning on that thread — happy to be overruled.
- **Bundle truncation.** Confirmed with a fixture rather than left theoretical: malformed-then-good parses **0** certificates. Needs a sentinel across the C boundary, so leaving it as follow-up work.
- **Test gap.** Agreed, but it belongs with that follow-up — a test here could only assert the truncation, which the fix then inverts.
- **`certs.scan_no_extension`.** Agreed this is the root cause the PR only mitigates, and agreed it is separate from this change.

## Verification after the changes

Linked `certs.c` against a driver under ASan: 192-cert system CA bundle parses clean with no NULL arrays and every `key_value` even-length; EC certs now carry their key; good cert -> 1, malformed -> 0 without crashing. `-Wall -Wextra` unchanged at 3 warnings, all pre-existing on lines 73 and 259.

nettrino (approved):
Reviewed revision 37ed84b. The malformed-public-key crash fix and regression test are sound, and the latest changes address the NULL-array handling, empty-key indexing, extension-name guards, and changelog corrections raised in the previous review. All three CI checks pass.

The remaining bundle-truncation behavior, incomplete allocation-failure handling, and broad scanning of extensionless files are nonblocking follow-ups. Please also align the PR description with the corrected changelog claims about EC metadata and BIO_all().

Approval is based on code review and the existing CI results; tests were not rerun locally.

